PR #5, opened 2019-05-09
Define cloud volume structures

jvinod:
Signed-off-by: Vinod <jvinod@portworx.com>

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -1,0 +1,84 @@
+package api
+
+import (
+	"time"
+)
+
+// CloudVolumeSpec defines volume spec.
+type CloudVolumeSpec struct {
+	// VolumeType the volume type (e.g. GP2 on AWS)
+	VolumeType *string `type:"string" enum:"VolumeType"`
+	// Size in GiBs.
+	Size *int64 `type:"integer"`
+	// Availability zone in which to create the volume.
+	AvailabilityZone *string `type:"string" required:"true"`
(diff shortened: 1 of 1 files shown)


api/types.go
```diff
@@ -0,0 +1,87 @@
...
+	InstanceID *string `locationName:"instanceID" type:"string"`
+	// State the attachment state of the volume.
+	State *string `locationName:"status" type:"string" enum:"VolumeAttachmentState"`
+	// VolumeID unique identifier for volume.
+	VolumeID *string `locationName:"volumeID" type:"string"`
+}
+
+// CloudVolume runtime status of CloudVolume.
+type CloudVolume struct {
+	// VolumeID unique identifier for the volume.
```
adityadani: Name should be a field here as well.

api/types.go
```diff
@@ -0,0 +1,87 @@
...
+	// InstanceID unique instance identifier.
+	InstanceID *string `locationName:"instanceID" type:"string"`
+	// State the attachment state of the volume.
+	State *string `locationName:"status" type:"string" enum:"VolumeAttachmentState"`
+	// VolumeID unique identifier for volume.
+	VolumeID *string `locationName:"volumeID" type:"string"`
+}
+
+// CloudVolume runtime status of CloudVolume.
+type CloudVolume struct {
```
piyush-nimbalkar: We should just have include the CloudVolumeSpec in here instead of redefining the fields here.

PR #6, opened 2019-05-10
 Add support for AWS and GCE instance operations

harsh-px:


cloudops.go
```diff
@@ -82,10 +82,12 @@ type InstanceInfo struct {
...
 	InstanceID() string
-	// Inspect instance identified by ID.
-	InspectIntance(ID string) (*InstanceInfo, error)
-	// InspectInstanceGroup returns instanceGroupInfo matching labels.
-	InspectInstanceGroup(labels map[string]string) (*InstanceInfo, error)
+	// InspectSelf inspects the node where the code is invoked. Hence this assumes
+	// this is invoked from within a cloud instance
+	InspectSelf() (*InstanceInfo, error)
+	// InspectSelfInstanceGroup inspects the instance group within which the current
+	// instance resides
```
jvinod: There is no need for the assumption that it is invoked from the same instance.   Take the InstanceID as a parameter to allow cloudops to be run as n API server from _an_ instance that has the _right_ privileges.

harsh-px: Done. Updated the compute APIs to

```
        // InspectSelf inspects the node where the code is invoked.
        InspectInstance(instanceID string) (*InstanceInfo, error)
        // InspectSelfInstanceGroupForInstance inspects the instance group to which
        // the cloud instance with given ID belongs
        InspectInstanceGroupForInstance(instanceID string) (*InstanceGroupInfo, error)
```

Used InspectInstanceGroupForInstance since our use case is to inspect the group by instance instead of a generic label query.

cloudops.go
```diff
@@ -82,10 +82,11 @@ type InstanceInfo struct {
 type Compute interface {
 	// InstanceID of instance where command is executed.
 	InstanceID() string
-	// Inspect instance identified by ID.
-	InspectIntance(ID string) (*InstanceInfo, error)
-	// InspectInstanceGroup returns instanceGroupInfo matching labels.
-	InspectInstanceGroup(labels map[string]string) (*InstanceInfo, error)
+	// InspectSelf inspects the node where the code is invoked.
```
jvinod: Comment needs to be updated

aws/aws.go
```diff
@@ -211,11 +226,83 @@ func (s *ec2Ops) waitAttachmentStatus(
...
+func (s *awsOps) InspectInstance(instanceID string) (*cloudops.InstanceInfo, error) {
+	inst, err := DescribeInstanceByID(s.ec2, instanceID)
+	if err != nil {
+		return nil, err
+	}
 
-func (s *ec2Ops) InstanceID() string { return s.instance }
+	instInfo := &cloudops.InstanceInfo{
+		CloudResourceInfo: cloudops.CloudResourceInfo{
+			Name:   instanceID,
```
adityadani: You can get the human readable instance Name from the tags
inst.Tags["Name"] will have the actual name of the instance, if it exists. We can resort to ID if there is no Name set.

aws/aws.go
```diff
@@ -211,11 +226,83 @@ func (s *ec2Ops) waitAttachmentStatus(
...
+				AutoscalingEnabled: true,
+				Min:                group.MinSize,
+				Max:                group.MaxSize,
+			}
+
+			return retval, nil
+		}
+	}
+
+	return nil, fmt.Errorf("instance doesn't belong to an instance group")
```
adityadani: this should be a well defined error instead of fmt.Errorf, so that the callers can decide to ignore if they want.

gce/gce.go
```diff
@@ -73,31 +82,161 @@ func NewClient() (cloudops.Ops, error) {
...
+				return nil, err
+			}
+		}
+	}
+
+	if len(gkeClusterName) == 0 ||
+		len(instanceTemplate) == 0 ||
+		len(clusterLocation) == 0 ||
+		len(kubeLabels) == 0 {
+		return nil, fmt.Errorf("API is currently only supported on the GKE platform")
```
adityadani: maybe a well defined error here as well.

harsh-px: done

aws/aws.go
harsh-px: done

aws/aws.go
harsh-px: added

PR #7, opened 2019-05-17
[WIP] Added SetCountForInstanceGroup() API and implementation for GCE

vinayakshnd:


test/cloudops.go
```diff
@@ -59,6 +59,12 @@ func compute(t *testing.T, driver cloudops.Ops) {
 	groupInfo, err := driver.InspectInstanceGroupForInstance(instanceID)
 	require.NoError(t, err, "failed to inspect instance group")
 	require.NotNil(t, groupInfo, "got nil instance group info from inspect")
+
+	err = driver.SetCountForInstanceGroup(groupInfo, int64(4))
+	require.NoError(t, err, "failed to inspect instance group")
```
harsh-px: Ignore if the driver returns not supported error

Also the message is incorrect

cloudops.go
```diff
@@ -48,6 +48,9 @@ type Compute interface {
 	// InspectInstanceGroupForInstance inspects the instance group to which the
 	// cloud instance with given ID belongs
 	InspectInstanceGroupForInstance(instanceID string) (*InstanceGroupInfo, error)
+	// SetCountForInstanceGroup sets desired count per availability zone
+	// for given instance group
+	SetCountForInstanceGroup(instanceGroupInfo *InstanceGroupInfo, count int64) error
```
harsh-px: Don't take the whole instance group info. Have the API specifically take in only what's needed

test/cloudops.go
vinayakshnd: @harsh-px updated error msg.
Any idea how we can avoid assertion failures for only certain kind of errors here?

cloudops.go
vinayakshnd: @harsh-px updated this API to just accept `instanceID` as input.
Also refactored both `InspectInstanceGroupForInstance ()` and `SetCountForInstanceGroup ()` to use common helper function `getGkeClusterDetails()`.

vinayakshnd:
@harsh-px updated PR with review comments

cloudops.go
```diff
@@ -48,6 +48,9 @@ type Compute interface {
 	// InspectInstanceGroupForInstance inspects the instance group to which the
 	// cloud instance with given ID belongs
 	InspectInstanceGroupForInstance(instanceID string) (*InstanceGroupInfo, error)
```
harsh-px: Can u also add support in the InspectInstanceGroupForInstance API to return current size. Just add the GKE implementation for now. We will add KOPS when we start on the KOPS effort.

gce/gce.go
```diff
@@ -673,6 +643,82 @@ func (s *gceOps) describeinstance() (*compute.Instance, error) {
 	return s.computeService.Instances.Get(s.inst.project, s.inst.zone, s.inst.name).Do()
 }
 
+func (s *gceOps) getGkeClusterDetails(instanceID string) (string, string, string, map[string]string, error) {
```
harsh-px: GKE instead of Gke

test/cloudops.go
```diff
@@ -59,6 +60,12 @@ func compute(t *testing.T, driver cloudops.Ops) {
 	groupInfo, err := driver.InspectInstanceGroupForInstance(instanceID)
 	require.NoError(t, err, "failed to inspect instance group")
 	require.NotNil(t, groupInfo, "got nil instance group info from inspect")
+
+	err = driver.SetCountForInstanceGroup(instanceID, int64(3))
+	require.NoError(t, err, "failed to set node count for instance group")
+	// TODO: Once current node count is tracked in instanceGroupInfo,
```
harsh-px: Add commented above, let's add the current node count support as this will be required by users of this API before invoking the set.

vinayakshnd:
Closing this PR in favour of new PR #9

PR #8, opened 2019-05-20
Add support for exponential backoff

adityadani:
- Added a wrapper interface over cloudops, that provides backoff mechanisms
- Exponential backoff is one such implementation.
- Each cloud provider has a callback function that checks if an exponential backoff is needed.

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/libopenstorage/cloudops"
+	"github.com/libopenstorage/cloudops/backoff"
 	"github.com/libopenstorage/cloudops/unsupported"
 	"github.com/portworx/sched-ops/task"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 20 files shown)


azure/azure.go
```diff
@@ -739,3 +739,25 @@ func lunToBlockDevPath(lun int32) (string, error) {
...
 	return devPath, nil
 }
+
+func isExponentialError(err error) bool {
+	// Got the list of error codes from here
+	// https://docs.microsoft.com/en-us/rest/api/storageservices/common-rest-api-error-codes
+
+	azureCodes := map[int]struct{}{
+		int(429): {},
+		int(503): {},
```
piyush-nimbalkar: Let's remove 503 as discussed.

backoff/exponential.go
```diff
@@ -0,0 +1,407 @@
...
+
+func (e *exponentialBackoff) Name() string {
+	return "exponential-backoff"
+}
+
+func (e *exponentialBackoff) handleError(origErr error, msg string) (bool, error) {
+	if origErr != nil {
+		if e.isExponentialError(origErr) {
+			// do an exponential backoff
+			logrus.Errorf("%v Retrying after a backoff...")
```
piyush-nimbalkar: nit: remove unused `%v` and `...`

PR #10, opened 2019-05-24
GCE: Use WaitForOpCompletion to check if any GCE API has completed.

adityadani:
- All GCE APIs are asynchronous. Every API returns an Operation object.
- Query the operation object to fetch the current status of the operation.
- The operation object also returns errors and error codes if any.

### DIFF
--- a/gce/gce.go
+++ b/gce/gce.go
@@ -40,6 +40,7 @@ const (
 	kubeLabelsKey           = "kube-labels"
 	nodePoolKey             = "cloud.google.com/gke-nodepool"
 	instanceTemplateKey     = "instance-template"
+	doneStatus              = "DONE"
 )
 
 type gceOps struct {
@@ -265,8 +266,11 @@ func (s *gceOps) ApplyTags(
 		Labels:           currentLabels,
 	}
 
-	_, err = s.computeService.Disks.SetLabels(s.inst.project, s.inst.zone, d.Name, rb).Do()
-	return err
+	operation, err := s.computeService.Disks.SetLabels(s.inst.project, s.inst.zone, d.Name, rb).Do()
+	if err != nil {
+		return err
+	}
+	return s.waitForOpCompletion("disk.ApplyTags", s.inst.zone, operation)
 }
 
 func (s *gceOps) Attach(diskName string) (string, error) {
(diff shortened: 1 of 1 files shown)


gce/gce.go
```diff
@@ -793,6 +823,58 @@ func (s *gceOps) waitForAttach(
...
+				if gErr, ok := err.(*googleapi.Error); ok {
+					if gErr.Code == int(404) {
+						// operation does not exist
+						return nil, false, nil
+					}
+				}
+				return nil, true, fmt.Errorf("failed to query gce operation %v for %v: %v", operation.Name, cloudopsOperationName, err)
+			}
+
+			if op == nil || op.Status != doneStatus {
```
piyush-nimbalkar: Even for failed operations, the status is DONE?

gce/gce.go
```diff
@@ -598,8 +625,11 @@ func (s *gceOps) Snapshot(
 }
 
 func (s *gceOps) SnapshotDelete(snapID string) error {
-	_, err := s.computeService.Snapshots.Delete(s.inst.project, snapID).Do()
-	return err
+	operation, err := s.computeService.Snapshots.Delete(s.inst.project, snapID).Do()
+	if err != nil {
+		return err
+	}
+	return s.waitForOpCompletion("SnapshotDelete", s.inst.zone, operation)
```
piyush-nimbalkar: *snapshot.Delete?

gce/gce.go
```diff
@@ -580,11 +603,15 @@ func (s *gceOps) Snapshot(
 		Name: fmt.Sprintf("snap-%d%02d%02d", time.Now().Year(), time.Now().Month(), time.Now().Day()),
 	}
 
-	_, err := s.computeService.Disks.CreateSnapshot(s.inst.project, s.inst.zone, disk, rb).Do()
+	operation, err := s.computeService.Disks.CreateSnapshot(s.inst.project, s.inst.zone, disk, rb).Do()
 	if err != nil {
 		return nil, err
 	}
 
+	if opErr := s.waitForOpCompletion("disk.Snapshot", s.inst.zone, operation); opErr != nil {
```
piyush-nimbalkar: *disk.CreateSnapshot?

gce/gce.go
adityadani: Yes that operation status is set to DONE and the error is updated if any.

PR #11, opened 2019-05-24
Fixes in azure provider

piyush-nimbalkar:
- Check if the disk is attached on remote node before trying to attach
- If disk is attached locally avoid the attach call
- Use multiple Get calls instead of List in Azure Inspect Disks
    - The request limit for List calls is quite low compared to the Get calls.
      List calls limit is 240/3min and 1900/30min
      Get calls limit is ~6000/3min and ~36000/30min
    - Using ListByResourceGroup instead of just List to scope the disk to
      the resource group. Also observed that the List call does not return
      a `Retry-after` header when returning 429, which the other does. This
      helps the SDK to retry after the said time instead of doing it immediately.
-  Adding Azure error AttachDiskWhileBeingDetached for exponential backoff.
    - Returning all errors as it is so the exponential error checker finds the original error.
- Detach a disk from azure vm if attach fails with AttachDiskWhileBeingDetached. When multiple VMs try to attach the same disk, Azure fails the calls but shows the disk attached to it. Any subsequent call returns AttachDiskWhileBeingDetached. To avoid this we try to do a detach prevent next attach from failing.

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -307,22 +307,31 @@ func (a *azureOps) FreeDevices(
 }
 
 func (a *azureOps) Inspect(diskNames []*string) ([]interface{}, error) {
-	allDisks, err := a.getDisks(nil)
-	if err != nil {
-		return nil, err
-	}
-
 	var disks []interface{}
-	for _, id := range diskNames {
-		if d, ok := allDisks[*id]; ok {
-			disks = append(disks, d)
-		} else {
-			return nil, cloudops.NewStorageError(
-				cloudops.ErrVolNotFound,
-				fmt.Sprintf("disk %s not found", *id),
-				a.instance,
-			)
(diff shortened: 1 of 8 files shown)


azure/azure.go
```diff
@@ -379,6 +416,16 @@ func (a *azureOps) Enumerate(
...
 
 func (a *azureOps) DevicePath(diskName string) (string, error) {
+	if _, err := a.checkDiskAttachmentStatus(diskName); err != nil {
+		return "", err
+	}
+	return a.devicePath(diskName)
+}
+
+// checkDiskAttachmentStatus returns the disk without any error if it is already
+// attached the Ops instance. It will return errors if the disk is not attached
```
adityadani: nit: attached to the Ops instance

PR #12, opened 2019-05-31
Fix nil panic when accessing autoscaling

harsh-px:

### DIFF
--- a/gce/gce.go
+++ b/gce/gce.go
@@ -219,10 +219,13 @@ func (s *gceOps) InspectInstanceGroupForInstance(instanceID string) (*cloudops.I
 					Zone:   s.inst.zone,
 					Region: s.inst.region,
 				},
-				Zones:              zones,
-				AutoscalingEnabled: nodePool.Autoscaling.Enabled,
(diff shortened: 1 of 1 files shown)


gce/gce.go
```diff
@@ -219,10 +219,13 @@ func (s *gceOps) InspectInstanceGroupForInstance(instanceID string) (*cloudops.I
...
-				Zones:              zones,
-				AutoscalingEnabled: nodePool.Autoscaling.Enabled,
-				Min:                &nodePool.Autoscaling.MinNodeCount,
-				Max:                &nodePool.Autoscaling.MaxNodeCount,
+				Zones: zones,
+			}
+
+			if nodePool.Autoscaling != nil {
+				retval.AutoscalingEnabled = nodePool.Autoscaling.Enabled
+				retval.Min = &nodePool.Autoscaling.MinNodeCount
```
adityadani: do the same for min and max node count

harsh-px: min and max node count are `int64`'s. So as long as `if nodePool.Autoscaling != nil {`, we are safe to access them.

PR #13, opened 2019-06-07
Lookup AWS ephemeral devices when detecting free devices

harsh-px:
Also 2 more changes
-  handle attachment failures due to devices in use
- use reversed list of free devices to increase probability of using devices not used by other components

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -107,7 +107,7 @@ func NewClient() (cloudops.Ops, error) {
 }
 
 // nvmeInstanceTypes are list of instance types whose EBS volumes are exposed as NVMe block devices
-var nvmeInstanceTypes = []string{"c5", "c5d", "i3.metal", "m5", "m5d", "r5", "r5d", "z1d"}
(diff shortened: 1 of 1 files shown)


aws/aws.go
```diff
@@ -766,23 +804,35 @@ func (s *awsOps) Attach(volumeID string) (string, error) {
...
-		return "", err
+
+	for _, device := range devices {
+		req := &ec2.AttachVolumeInput{
+			Device:     &device,
+			InstanceId: &s.instance,
+			VolumeId:   &volumeID,
+		}
+		if _, err := s.ec2.AttachVolume(req); err != nil {
+			if strings.Contains(err.Error(), "is already in use") {
```
adityadani: nit: make sure this error string does not match with error returned when attach command fails because volume is attached on another node/already in use.

harsh-px: Just tested that and we are good. If the volume is being used by another instance, it returns

```
ERRO[0000] failed to attach: VolumeInUse: vol-049cbf86e507578d2 is already attached to an instance
        status code: 400, request id: 0a85d578-0422-46a1-8f1c-50364a7679ba
```

PR #15, opened 2019-06-13
[PTX-1454] Changes required for torpedo GKE ASG test cases

vinayakshnd:


gce/gce.go
```diff
@@ -688,6 +694,45 @@ func (s *gceOps) SetInstanceGroupSize(instanceGroupID, instanceGroupLocation str
...
+	if err != nil {
+		return 0, err
+	}
+
+	nodeCount := int64(0)
+	for _, instanceGroupURL := range nodePool.InstanceGroupUrls {
+
+		nodeGrpName := strings.TrimSpace(filepath.Base(instanceGroupURL))
+
+		temp := strings.SplitAfter(instanceGroupURL, "zones")[1]
```
harsh-px: do a check here for the size of the slice after split. For any reason, if zones is not present in the instanceGroupURL, this will panic.

gce/gce.go
```diff
@@ -688,6 +694,45 @@ func (s *gceOps) SetInstanceGroupSize(instanceGroupID, instanceGroupLocation str
...
+		zone := strings.Split(temp, "/")[1]
```
harsh-px: https://golang.org/pkg/path/#Base might be a better fit here instead of split and indexing.

gce/gce.go
```diff
@@ -615,35 +619,37 @@ func (s *gceOps) RemoveTags(
...
 		return err
 	}
 
+	var cluster *container.Cluster
 	if zonalCluster {
-		_, err = s.containerService.Projects.Zones.Clusters.NodePools.SetSize(s.inst.project,
-			instanceGroupLocation,
-			clusterName,
-			instanceGroupID,
+		_, err = s.containerService.Projects.Zones.Clusters.NodePools.SetSize(
```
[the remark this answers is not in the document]
vinayakshnd: @harsh-px ,
SetSize() returns [Operation](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.operations#Operation) and error.
I am not using it, so not storing it.

gce/gce.go
vinayakshnd: Hi @harsh-px ,

https://golang.org/pkg/path/#Base will be useful while getting last element of the URL. Hence used it here to extract `nodeGrpName`

```suggestion
		nodeGrpName := strings.TrimSpace(filepath.Base(instanceGroupURL))
```

But for extracting zone from bellow url, I am not sure we can use it.
https://www.googleapis.com/compute/v1/projects/portworx-eng/zones/us-central1-a/instanceGroupManagers/gke-torpedo-tp-2-1-2-gke-default-pool-fbff8a1b-grp

gce/gce.go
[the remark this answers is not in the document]
harsh-px: In that case, looks like the correct implementation would be to check the returned Operation object for status first. So you will first have a retry loop to check if the operation succeeded and then perform the size check.

PR #16, opened 2019-06-27
[PTX-1608] Added DeleteInstance() API interface

vinayakshnd:
Added DeleteInstance() API interface and its implementation for GCE

### DIFF
--- a/cloudops.go
+++ b/cloudops.go
@@ -43,6 +43,8 @@ type InstanceInfo struct {
 
 // Compute interface to manage compute instances.
 type Compute interface {
+	// DeleteInstance deletes the instance
+	DeleteInstance(instanceID string, zone string) error
 	// InstanceID of instance where command is executed.
 	InstanceID() string
 	// InspectInstance inspects the node with the given instance ID
--- a/cloudops.go
+++ b/cloudops.go
@@ -54,9 +54,11 @@ type Compute interface {
 	InspectInstanceGroupForInstance(instanceID string) (*InstanceGroupInfo, error)
 	// SetInstanceGroupSize sets desired node count per availability zone
 	// for given instance group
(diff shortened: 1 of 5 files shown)


gce/gce.go
```diff
@@ -382,6 +382,131 @@ func (s *gceOps) DeleteFrom(id, _ string) error {
 	return s.Delete(id)
 }
 
+func (s *gceOps) DeleteInstance(instanceID string) error {
+
+	if s.inst.clusterName == "" || s.inst.clusterLocation == "" {
+		// We delete an instance only if its part of GKE. Not GCE.
+		return fmt.Errorf("running on non-GKE environment. Not deleting instance [%v]", instanceID)
```
harsh-px: The delete instance method doesn't need to add a "cluster" restriction and should work on plain GCP itself.

So take in the instance zone as an argument. 

To keep the APIs consistent, do the same for InspectInstance().

vinayakshnd: Hi @harsh-px , we can pass instance zone to DeleteInstance() and InspectInstance() but the problem is the calling API needs to know the zone of that instance then.  Right now there is no way to find out zone of an instance just by its ID. (Instance residing in zone other than where gke SDK is running.)

cloudops.go
```diff
@@ -43,6 +43,8 @@ type InstanceInfo struct {
 
 // Compute interface to manage compute instances.
 type Compute interface {
+	// DeleteInstance deletes the instance
+	DeleteInstance(instanceID string, zone string, timeout time.Duration) error
 	// InstanceID of instance where command is executed.
 	InstanceID() string
 	// InspectInstance inspects the node with the given instance ID
```
harsh-px: Add a TODO above to fix the `InspectInstance` API to take in a zone too.

PR #18, opened 2019-08-01
Update azure VM even if the disk is not found to fix stuck detach issue

piyush-nimbalkar:
This is a known issue from Azure, and updating VM is the suggested workaround.
https://github.com/andyzhangx/demo/blob/master/issues/azuredisk-issues.md#18-detach-azure-disk-make-vm-run-into-a-limbo-state

Signed-off-by: Piyush Nimbalkar <piyush@portworx.com>

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -265,11 +265,26 @@ func (a *azureOps) detachInternal(diskName, instance string) error {
 		a.resourceGroupName,
 		diskName,
 	)
-	if err != nil {
+	if derr, ok := err.(autorest.DetailedError); ok {
+		if code, ok := derr.StatusCode.(int); !ok {
+			return err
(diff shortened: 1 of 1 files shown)


azure/azure.go
```diff
@@ -284,11 +299,6 @@ func (a *azureOps) detachInternal(diskName, instance string) error {
 		newDataDisks = append(newDataDisks, d)
 	}
 
-	// If the disk is not attached, return success
```
adityadani: If we are getting rid of this check, then we need to add an equivalent check after updateDataDisks assuming in a normal scenario when (dataDisks == newDataDisks), the updateDataDisks API fails

piyush-nimbalkar: The API does not fail if you update the VM without any updates.

PR #19, opened 2019-08-02
Azure Storage Distribution

adityadani:
The azure storage distribution algorithm provides an optimum
  storage distribution strategy for a given set of following inputs:
  - Requested IOPS from cloud storage.
  - Minimum capacity for the whole cluster.
  - Number of zones in the cluster.
  - Number of instances in the cluster.
  - An azure storage distribution matrix.

  It uses the following algorithm:

  1. For the requested instances per zone and total zone count
    2. Calculate the number of total storage instances
    3. Divide the minimum capacity with total instances to get "PerNodeCapacity"
    4. Check with the matrix if PerNodeCapacity satisfies the requested IOPS for any drive
    5. Generate a list of candidates (drive types and sizes) that satisfy the requirements
    6. Sort the candidates by closest to requested IOPS such that the first element would
       be closest to the requested IOPS
    7. Remove the candidates whose IOPS are greater than requested by a factor of 1000 (constant could be changed?)
    8. If there are still multiple candidates use priority.
    9. If no candidates were found in this process reduce the requested instances per zone by 1 and go to Step 1.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -570,6 +570,14 @@
   version = "v1.20.1"
 
 [[projects]]
+  digest = "1:4d2e5a73dc1500038e504a8d78b986630e3626dc027bc030ba5c75da257cdb96"
+  name = "gopkg.in/yaml.v2"
+  packages = ["."]
+  pruneopts = "UT"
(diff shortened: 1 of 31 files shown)


adityadani:
Closing this PR. Created a new one with an updated algorithm #20

PR #21, opened 2019-08-07
Add StorageDecisionMatrix yaml parser.

adityadani:
Signed-off-by: Aditya Dani <aditya@portworx.com>

### DIFF
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -1,0 +1,89 @@
+package parser
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/libopenstorage/cloudops"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testYamlFilePath     = "/tmp/cloudops-test.yaml"
+	existingYamlFilePath = "testspecs/test.yaml"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/parser/parser_test.go
```diff
@@ -0,0 +1,40 @@
...
+			cloudops.StorageDecisionMatrixRow{
+				IOPS:         uint32(2000),
+				MinSize:      uint64(200),
+				MaxSize:      uint64(400),
+				InstanceType: "bar",
+			},
+		},
+	}
+	p := NewStorageDecisionMatrixParser()
+	err := p.MarshalToYaml(&inputMatrix, testYamlFilePath)
```
harsh-px: this test doesn't validate if this works with a given yaml file since we are marshalling the struct into yaml and then unmarshalling the same file.

We should rather be starting the test by calling `UnmarshalFromYaml` on https://github.com/libopenstorage/cloudops/blob/master/azure/testspecs/azure-storage-decision-matrix.yaml.

adityadani: done!

Found an issue with the azure test spec after adding the test :)

PR #24, opened 2019-08-09
Add Expand API to Storage CloudOps interface.

adityadani:
Added Expand API to Cloudops Disk interface.  
- Currently implemented only for Azure and AWS
- Added a UT to test Expand API.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -25,6 +25,7 @@ import (
 	awscredentials "github.com/libopenstorage/secrets/aws/credentials"
 	"github.com/portworx/sched-ops/task"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


aws/aws.go
```diff
@@ -843,6 +844,66 @@ func (s *awsOps) detachInternal(volumeID, instanceName string) error {
...
+func (s *awsOps) Expand(
+	volumeID string,
+	newSizeInGiB uint64,
+) (uint64, error) {
+	vol, err := s.refreshVol(&volumeID)
+	if err != nil {
+		return 0, err
+	}
+	currentSizeInGiB := uint64(*vol.Size)
+	if uint64(*vol.Size) >= newSizeInGiB {
```
harsh-px: use `currentSizeInGiB` instead of `uint64(*vol.Size)`

Also we must fail the Expand request with error that volume is already at a larger size than the requested size. The caller should decide what to do with the error.

azure/azure.go
```diff
@@ -326,6 +326,57 @@ func (a *azureOps) DeleteFrom(diskName, _ string) error {
...
+	if err != nil {
+		return 0, err
+	}
+
+	if disk.DiskProperties == nil || disk.DiskProperties.DiskSizeGB == nil {
+		return 0, fmt.Errorf("disk properties of (%v) is nil", diskName)
+	}
+
+	if *disk.DiskProperties.DiskSizeGB >= int32(newSizeInGiB) {
+		// no work is needed
```
harsh-px: same as AWS. The API should return an error here.

azure/azure.go
```diff
@@ -326,6 +326,57 @@ func (a *azureOps) DeleteFrom(diskName, _ string) error {
...
+		return uint64(*disk.DiskProperties.DiskSizeGB), nil
+	}
+	oldSizeInGiB := uint64(*disk.DiskProperties.DiskSizeGB)
+	// Azure resizes in chunks of GiB even if the disk properties variable is DiskSizeGB
+	newSizeInGiBInt32 := int32(newSizeInGiB)
+	disk.DiskProperties.DiskSizeGB = &newSizeInGiBInt32
+
+	ctx := context.Background()
+	future, err := a.disksClient.CreateOrUpdate(
```
piyush-nimbalkar: nit: Just call Update here instead of CreateOrUpdate. Although it doesn't matter because you are doing a Get before so it won't create a new disk anyway.

azure/azure.go
```diff
@@ -344,8 +344,9 @@ func (a *azureOps) Expand(
 	}
 
 	if *disk.DiskProperties.DiskSizeGB >= int32(newSizeInGiB) {
-		// no work is needed
-		return uint64(*disk.DiskProperties.DiskSizeGB), nil
+		return 0, cloudops.NewStorageError(cloudops.ErrDiskGreaterOrEqualToExpandSize,
+			fmt.Sprintf("disk is already has a size: %d greater than or equal "+
```
piyush-nimbalkar: nit: *current disk size is greater than or equal ...

PR #28, opened 2019-09-05
Add comments about UDEV disk rules for GCE

rodrigc:


rodrigc:
I discovered this with a custom image provided by RedHat for Openshift 4.2.

See:
https://bugzilla.redhat.com/show_bug.cgi?id=1748638

PR #29, opened 2019-09-05
Add comments about UDEV disk rules for Azure

rodrigc:


rodrigc:
I found this when trying to use a custom-provisioned image provided by RedHat for Openshift:

https://bugzilla.redhat.com/show_bug.cgi?id=1747575

PR #36, opened 2019-10-24
Add lower sized disks for Azure Standard_LRS disk types.

adityadani:


azure/storagemanager/azure_test.go
```diff
@@ -98,10 +98,10 @@ func storageDistribution(t *testing.T) {
 			response: &cloudops.StorageDistributionResponse{
 				InstanceStorage: []*cloudops.StoragePoolSpec{
 					&cloudops.StoragePoolSpec{
-						DriveCapacityGiB: 113,
+						DriveCapacityGiB: 37,
```
adityadani: StandardLRS disks now start from 32 instead of 64

azure/storagemanager/azure_test.go
```diff
@@ -269,8 +270,8 @@ func storageDistribution(t *testing.T) {
 			response: &cloudops.StorageDistributionResponse{
 				InstanceStorage: []*cloudops.StoragePoolSpec{
 					&cloudops.StoragePoolSpec{
-						DriveCapacityGiB: 111,
-						DriveType:        "Standard_LRS",
+						DriveCapacityGiB: 128,
+						DriveType:        "StandardSSD_LRS",
```
adityadani: Modify the test to pass in a DriveType and expect the same drive type. The same configuration was being handled in TestCase 2

azure/storagemanager/azure_test.go
```diff
@@ -305,25 +306,7 @@ func storageDistribution(t *testing.T) {
 				InstancesPerZone: 3,
 				ZoneCount:        3,
 			},
-			response: &cloudops.StorageDistributionResponse{
```
adityadani: Change the testcase so that it does what the comment for the test case says.

PR #37, opened 2019-11-01
Storage Distribution Algo support for AWS.

adityadani:
The changes are divided into 3 commits

1st commit - New yaml file for AWS. Changes to existing azure yaml file.
```
- The aws yaml is generated and not hand written.
- Introduced min and max iops in a row. Modified azure yaml to include this new change.
```

2nd commit - Code to generate AWS yaml

3rd commit - Support for AWS storage distribution algorithm
```
- Change the existing implementation of SortByIOPS and FilterByIOPS to consider
  an IOPS range.
- In case of AWS, the size of the disk dictates the IOPS. Moved the logic of
  determining the IOPS of a StoragePool from the common algorithm to per cloud based logic.
- ResizeDisk: For resize disk instead of considering only one row which fits the current drive
  size and iops. Choose the all the rows which allow the existing IOPS and the drive size.
```


vsphere/storagemanager/vsphere.go
```diff
@@ -23,12 +23,37 @@ func newVsphereStorageManager(
 func (a *vsphereStorageManager) GetStorageDistribution(
 	request *cloudops.StorageDistributionRequest,
 ) (*cloudops.StorageDistributionResponse, error) {
-	return storagedistribution.GetStorageDistribution(request, a.decisionMatrix)
+	response := &cloudops.StorageDistributionResponse{}
+	for _, userRequest := range request.UserStorageSpec {
+		// for for request, find how many instances per zone needs to have storage
+		// and the storage spec for each of them
+		instStorage, instancePerZone, _, err :=
```
harsh-px: nit: rename to instancesPerZone

specs/decisionmatrix/aws.yaml
```diff
@@ -0,0 +1,3620 @@
+rows:
```
harsh-px: in the generated file, add a comment to the top that this file is generated automatically using the the make rule and do not edit manually

aws/storagemanager/spec/Makefile
```diff
@@ -0,0 +1,4 @@
+gen:
```
harsh-px: we should probably run the gen rule in travis and fail if the aws.yaml generated is different from the one checked in. Can be in a followup PR.

aws/storagemanager/aws_test.go
```diff
@@ -0,0 +1,708 @@
...
+						DriveCapacityGiB: 700,
+						DriveType:        "gp2",
+						DriveCount:       1,
+						IOPS:             2100,
+					},
+				},
+			},
+			expectedErr: nil,
+		},
+		/*{
```
harsh-px: is this commented deliberately?

cloud_storage_management.go
```diff
@@ -208,23 +210,41 @@ func (dm *StorageDecisionMatrix) FilterByDriveType(requestedDriveType string) *S
 	return dm
 }
 
-// FilterByIOPS filters out the rows which do not meet the requested IOPS.
+// FilterByMinIOPS filters out the rows whose minIOPS are greater requested IOPS.
```
harsh-px: change to `filters out the rows whose minIOPS are less than requested IOPS'

cloud_storage_management.go
```diff
@@ -237,6 +257,23 @@ func (dm *StorageDecisionMatrix) FilterByDriveSize(currentDriveSize uint64) *Sto
 	return dm
 }
 
+// FilterByDriveSize filters out the rows for which the current drive size is greater than
+// row's min size
```
harsh-px: add function comment to cover the `else if`

pkg/storagedistribution/storagedistribution.go
```diff
@@ -348,9 +312,10 @@ row_loop:
 
 	if int(rowIndex) == len(dm.Rows) {
 		// row_loop failed
-		return nil, 0, cloudops.ErrStorageDistributionCandidateNotFound
+		return nil, 0, nil, cloudops.ErrStorageDistributionCandidateNotFound
 	}
 
+	fmt.Println("$$$$ minCapacityPerZone: ", minCapacityPerZone, driveCount, driveSize)
```
harsh-px: remove

aws/storagemanager/aws_test.go
adityadani: this is the same test case which has been commented for all providers. The algorithm currently cannot handle adding disks to storage less nodes.

aws/storagemanager/spec/Makefile
adityadani: good idea. Will do.

pkg/storagedistribution/storagedistribution.go
adityadani: ah! this one got slipped in. Will fix

PR #41, opened 2019-11-16
 PWX-10524: handle cases when delta per disk is a float number.

trierra:
For instance 2.1 will be rounded up to 3, and sum of new disks capacity is always >= desired capacity


azure/storagemanager/azure_test.go
```diff
@@ -678,6 +678,54 @@ func storageUpdate(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			// ***** TEST: 14 instance is already at higher capacity than requested
```
harsh-px: fix test comments before commit

PR #43, opened 2019-12-10
when instances max drives, just print the max drive count in error

harsh-px:

### DIFF
--- a/pkg/storagedistribution/storagedistribution.go
+++ b/pkg/storagedistribution/storagedistribution.go
@@ -118,10 +118,17 @@ func AddDisk(
 		}
 	}
 
+	maxDriveCount := dm.Rows[0].InstanceMaxDrives
+	for _, row := range dm.Rows {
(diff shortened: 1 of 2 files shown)


pkg/storagedistribution/storagedistribution.go
```diff
@@ -118,10 +118,17 @@ func AddDisk(
 		}
 	}
 
+	maxDriveCount := dm.Rows[0].InstanceMaxDrives
+	for _, row := range dm.Rows {
```
piyush-nimbalkar: This may not be true for all instance types. Can you add a note here for future so we remember to modify this when we have specific instance types in the the rows?

adityadani: For each instance types we will have a separate row and a corresponding max drive count limit.
So we will need this even in that case right ?

adityadani: Basically we will have a drive count limit (openstorage/PX enforce) or (cloud) enforced.

PR #44, opened 2019-12-11
improve err messages when failing to find a drive size

harsh-px:

### DIFF
--- a/pkg/storagedistribution/storagedistribution.go
+++ b/pkg/storagedistribution/storagedistribution.go
@@ -3,6 +3,7 @@ package storagedistribution
 import (
 	"fmt"
 	"math"
+	"sort"
 
 	"github.com/libopenstorage/cloudops"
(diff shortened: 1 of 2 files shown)


pkg/storagedistribution/storagedistribution.go
```diff
@@ -111,10 +112,22 @@ func AddDisk(
 		}
 	}
 
+	driveRanges := make(map[string]string)
+	driveRangeStr := make([]string, 0)
+	for _, row := range dm.Rows {
```
piyush-nimbalkar: nit: Don't need to iterate twice. We can directly add the driveRangeStr slice, right?

harsh-px: wanted to avoid duplicates incase 2 rows have the same ranges.

piyush-nimbalkar: Got it.

PR #45, opened 2019-12-20
[PTX-1632] Added GetInstanceGroupSize() and SetInstanceGroupSize() API for Azure

vinayakshnd:


azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+// NewAzureClient initializes cloudops driver for azure based on environment
+// variables or based on instance metadata info available inside Azure VM
+func NewAzureClient() (cloudops.Ops, error) {
+
+	if onAzure() {
+		logrus.Info("Running on Azure IaaS VM")
+
+		client := &http.Client{}
+
+		req, _ := http.NewRequest("GET", metadataAPIEndpoint, nil)
```
harsh-px: nit: don't ignore err here

azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+		q.Add("api-version", metadataAPIVersion)
+		req.URL.RawQuery = q.Encode()
+
+		resp, err := client.Do(req)
+		if err != nil {
+			logrus.Errorf("Error occured while getting instance metadata from Azure Metadata API. Error:[%v]", err)
+			return nil, err
+		}
+
+		defer resp.Body.Close()
```
harsh-px: check if `resp.Body != nil` before calling Close()..

azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+		if resp.StatusCode != 200 {
+			return nil, fmt.Errorf("Error querying Azure metadata: Code %d returned for url %s", resp.StatusCode, req.URL)
+		}
+		respBody, _ := ioutil.ReadAll(resp.Body)
```
harsh-px: don't ignore err.

azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+			temp := resourceGroup.(string)
+			if len(strings.Split(temp, "_")) == 2 {
+				resourceGroupName = strings.Split(temp, "_")[1]
+				managedClusterName = strings.Split(temp, "_")[2]
+			}
+		}
+
+		if scalesetNametemp, exists = metadata[scaleSetNameKey]; exists {
+			scalesetName := scalesetNametemp.(string)
+			agentPoolName = strings.Split(scalesetName, "-")[1]
```
harsh-px: check length before indexing

azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+
+		var vmID, scalesetName, subscriptionID, resourceGroupName, managedClusterName, agentPoolName string
+		var scalesetNametemp, vmIDtemp, subscriptionIDtemp interface{}
+		var exists bool
+
+		if resourceGroup, ok := metadata[resourceGroupNameKey]; ok {
+			temp := resourceGroup.(string)
+			if len(strings.Split(temp, "_")) == 2 {
+				resourceGroupName = strings.Split(temp, "_")[1]
+				managedClusterName = strings.Split(temp, "_")[2]
```
harsh-px: unless i'm missing something, this will panic. `if len(strings.Split(temp, "_")) == 2` means where are 2 items in the list so you cannot index `strings.Split(temp, "_")[2]`.

azure/azure.go
```diff
@@ -46,11 +58,112 @@ var (
...
+		logrus.Errorf("Error occured while getting instance metadata from Azure Metadata API. Error:[%v]."+
+			"Probably not running on Azure IaaS VMs created by Azure Resource Manager.", err)
+		return false
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		logrus.Errorf("Error querying Azure metadata: Code %d returned for url %s", resp.StatusCode, req.URL)
+		return false
+	}
```
harsh-px: duplicate of snipped in `NewAzureClient()`. Put in common method and have it satisfy both use cases

azure/azure.go
```diff
@@ -114,6 +251,88 @@ func (a *azureOps) InstanceID() string {
 	return a.instance
 }
 
+func (a *azureOps) InspectInstance(instanceID string) (*cloudops.InstanceInfo, error) {
+
+	instInfo := &cloudops.InstanceInfo{
+		CloudResourceInfo: cloudops.CloudResourceInfo{
+			Name:   a.instance,
+			ID:     a.instance,
+			Zone:   "",
```
harsh-px: are both zone and region not populated by the azure API?

azure/azure.go
```diff
@@ -114,6 +251,88 @@ func (a *azureOps) InstanceID() string {
...
+	}
+	return instInfo, nil
+}
+
+func (a *azureOps) InspectInstanceGroupForInstance(instanceID string) (*cloudops.InstanceGroupInfo, error) {
+
+	ctx := context.Background()
+	agentPool, err := a.agentPoolsClient.Get(ctx, a.resourceGroupName, a.managedClusterName, a.agentPoolName)
+	if err != nil {
+		logrus.Errorf("Failed to get node pool details for [%s] of cluster [%v] in [%v] resource group."+
```
harsh-px: cloudops is a library and it should not be logging any errors. Add the same detail to `err` and just return that. Same for other places in the PR.

test/cloudops.go
```diff
@@ -156,6 +148,17 @@ func compute(t *testing.T, driver cloudops.Ops) {
...
+	instanceToDelete := os.Getenv("INSTANCE_TO_DELETE")
+	zoneOfInstanceToDelete := os.Getenv("INSTANCE_TO_DELETE_ZONE")
+	if instanceToDelete != "" {
+		err := driver.DeleteInstance(instanceToDelete, zoneOfInstanceToDelete, 5*time.Minute)
+		if _, ok := err.(*cloudops.ErrNotSupported); ok {
+			return
+		}
+		require.NoError(t, err, fmt.Sprintf("failed to delete instance [%v]. Error:[%v]", instanceToDelete, err))
+	} else {
+		logrus.Fatalf("Set INSTANCE_TO_DELETE environment variable")
```
harsh-px: looks like the test user has to create a dedicated vm to get deleted before running the test. This should be optional and should not fail other tests.

azure/azure.go
```diff
@@ -114,6 +251,88 @@ func (a *azureOps) InstanceID() string {
...
+	ctx := context.Background()
+	var cancel context.CancelFunc
+	if timeout > time.Nanosecond {
+		// If timeout is non-zero then
+		// wait infinitely for operation to complete
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	instaceGroupSize := int32(count)
```
harsh-px: typo in `instaceGroupSize`

azure/azure.go
```diff
@@ -114,6 +251,88 @@ func (a *azureOps) InstanceID() string {
...
+
+	ctx := context.Background()
+	agentPool, err := a.agentPoolsClient.Get(ctx, a.resourceGroupName, a.managedClusterName, a.agentPoolName)
+	if err != nil {
+		logrus.Errorf("Failed to get node pool details for [%s] of cluster [%v] in [%v] resource group."+
+			"Error: [%v] ", a.agentPoolName, a.managedClusterName, a.resourceGroupName, err)
+		return nil, err
+	}
+
+	// TODO: Currently multi-availability feature is in preview and
```
harsh-px: @piyush-nimbalkar should review this. Odd for azure API to not support zones and regions.

test/cloudops.go
harsh-px: actually looks like this was already existing. Just make the variable optional as part of this change.

azure/azure.go
vinayakshnd: @harsh-px / @piyush-nimbalkar , when I try to get agentPool details by using `a.agentPoolsClient.Get()` API call, I get bellow response.

For Availability Sets,
Screenshot 2020-01-07 at 2 08 14 PM
```
AgentPool details: [containerservice.AgentPool{Response:autorest.Response{Response:(*http.Response)(0xc00027ca20)}, ManagedClusterAgentPoolProfileProperties:(*containerservice.ManagedClusterAgentPoolProfileProperties)(0xc000128100), ID:(*string)(0xc0004d5550), Name:(*string)(0xc0004d5560), Type:(*string)(0xc0004d51e0)}]

Availability Zones: [(*[]string)(nil)]

agentPool.ManagedClusterAgentPoolProfileProperties: [&containerservice.ManagedClusterAgentPoolProfileProperties{Count:(*int32)(0xc0004c5568), VMSize:"Standard_DS3_v2", OsDiskSizeGB:(*int32)(0xc0004c556c), VnetSubnetID:(*string)(nil), MaxPods:(*int32)(0xc0004c55a4), OsType:"Linux", MaxCount:(*int32)(nil), MinCount:(*int32)(nil), EnableAutoScaling:(*bool)(nil), Type:"AvailabilitySet", OrchestratorVersion:(*string)(0xc0004d5540), ProvisioningState:(*string)(0xc0004d5530), AvailabilityZones:(*[]string)(nil)}]

agentPool.ManagedClusterAgentPoolProfileProperties.AvailabilityZones: [(*[]string)(nil)]

agentPool.EnableAutoScaling: [(*bool)(nil)]

agentPool.MaxCount: [<nil>], agentPool.MinCount: [<nil>]
```

For VMSS scale set,
Screenshot 2020-01-07 at 2 12 12 PM

```
AgentPool details: [containerservice.AgentPool{Response:autorest.Response{Response:(*http.Response)(0xc0001f02d0)}, ManagedClusterAgentPoolProfileProperties:(*containerservice.ManagedClusterAgentPoolProfileProperties)(0xc0001fa380), ID:(*string)(0xc0004dd660), Name:(*string)(0xc0004dd670), Type:(*string)(0xc0004dd390)}]

Availability Zones: [(*[]string)(nil)]
agentPool.ManagedClusterAgentPoolProfileProperties: [&containerservice.ManagedClusterAgentPoolProfileProperties{Count:(*int32)(0xc00003c088), VMSize:"Standard_DS3_v2", OsDiskSizeGB:(*int32)(0xc00003c08c), VnetSubnetID:(*string)(nil), MaxPods:(*int32)(0xc00003c0c4), OsType:"Linux", MaxCount:(*int32)(nil), MinCount:(*int32)(nil), EnableAutoScaling:(*bool)(nil), Type:"VirtualMachineScaleSets", OrchestratorVersion:(*string)(0xc0004dd650), ProvisioningState:(*string)(0xc0004dd640), AvailabilityZones:(*[]string)(nil)}]

agentPool.ManagedClusterAgentPoolProfileProperties.AvailabilityZones: [(*[]string)(nil)]

agentPool.EnableAutoScaling: [(*bool)(nil)]

agentPool.MaxCount: [<nil>], agentPool.MinCount: [<nil>]

```

vinayakshnd:
Hi @harsh-px addressed review comments.

azure/azure.go
piyush-nimbalkar: I just did this from azure cli (with debug to see what API it calls), which calls the same API as you are and it has zone information - 
```
az aks nodepool show --cluster-name piyush-test-mz -g piyush-test-mz --name nodepool1 --debug
```
```
GET /subscriptions/XXX/resourceGroups/piyush-test-mz/providers/Microsoft.ContainerService/managedClusters/piyush-test-mz/agentPools/nodepool1?api-version=2019-08-01

{
  "agentPoolType": "VirtualMachineScaleSets",
  "availabilityZones": [
    "1",
    "2",
    "3"
  ],
  "count": 0,
  "enableAutoScaling": true,
  "enableNodePublicIp": null,
  "id": "/subscriptions/XXXX/resourcegroups/piyush-test-mz/providers/Microsoft.ContainerService/managedClusters/piyush-test-mz/agentPools/nodepool1",
  "maxCount": 6,
  "maxPods": 110,
  "minCount": 3,
  "name": "nodepool1",
  "nodeTaints": null,
  "orchestratorVersion": "1.13.10",
  "osDiskSizeGb": 100,
  "osType": "Linux",
  "provisioningState": "Succeeded",
  "resourceGroup": "piyush-test-mz",
  "scaleSetEvictionPolicy": null,
  "scaleSetPriority": null,
  "type": "Microsoft.ContainerService/managedClusters/agentPools",
  "vmSize": "Standard_DS2_v2",
  "vnetSubnetId": null
}
```
The cluster I tried was a multi-zone aks cluster.
When an AKS cluster does not in AZs it is running in Availability Sets (AS) which is a different construct than AZ. 
A VM can run either in an AZ or AS. So we should populate the values if present else may be set an empty array instead of dummy value.
@harsh-px @vinayakshnd - How is the Zones field used in cloudops or it's user?

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+}
+
+func (a *azureOps) GetInstanceGroupSize(instanceGroupID string) (int64, error) {
+
+	ctx := context.Background()
+	agentPool, err := a.agentPoolsClient.Get(ctx, a.resourceGroupName, a.managedClusterName, instanceGroupID)
+	if err != nil {
+		return 0, err
+	}
+	return int64(*agentPool.Count), nil
```
piyush-nimbalkar: As this is a pointer, may be we should check that this is not null and return a default value with or without error in case it is nil.
The doc in the azure sdk says - 
```
// Count - Number of agents (VMs) to host docker containers. Allowed values must be in the range of 1 to 100 (inclusive). The default value is 1
```
@harsh-px any harm returning 1 if this is nil for some reason?

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+// SetInstanceGroupSize sets desired node count per availability zone
+// for given instance group
+func (a *azureOps) SetInstanceGroupSize(instanceGroupID string,
+	count int64,
+	timeout time.Duration) error {
+
+	ctx := context.Background()
+	var cancel context.CancelFunc
+	if timeout > time.Nanosecond {
+		// If timeout is non-zero then
```
piyush-nimbalkar: *If timeout is zero then ...

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+		// wait infinitely for operation to complete
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	instanceGroupSize := int32(count)
+	agentPoolProperties := containerservice.ManagedClusterAgentPoolProfileProperties{
```
piyush-nimbalkar: Does the update work properly when you are not sending the other original values in the AgentPool object?
Should we not do a get and only update the count and call CreateOrUpdate?

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+	}
+	result, err := a.agentPoolsClient.CreateOrUpdate(ctx, a.resourceGroupName,
+		a.managedClusterName,
+		a.agentPoolName,
+		agentPool)
+	if err != nil {
+		return fmt.Errorf("Failed to set sized for node group [%v] of managed cluster [%v]."+
+			"Error:[%v]", a.agentPoolName, a.managedClusterName, err)
+	}
+	result.WaitForCompletionRef(ctx, a.agentPoolsClient.Client)
```
piyush-nimbalkar: return error if WaitForCompletionRef returns an error

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+
+	instanceGroupSize := int32(count)
+	agentPoolProperties := containerservice.ManagedClusterAgentPoolProfileProperties{
+		Count: &instanceGroupSize,
+	}
+
+	agentPool := containerservice.AgentPool{
+		ManagedClusterAgentPoolProfileProperties: &agentPoolProperties,
+	}
+	result, err := a.agentPoolsClient.CreateOrUpdate(ctx, a.resourceGroupName,
```
piyush-nimbalkar: nit: this is a future and not actually the result

azure/azure.go
```diff
@@ -114,6 +241,87 @@ func (a *azureOps) InstanceID() string {
...
+			Region: "",
+		},
+	}
+	return instInfo, nil
+}
+
+func (a *azureOps) InspectInstanceGroupForInstance(instanceID string) (*cloudops.InstanceGroupInfo, error) {
+
+	ctx := context.Background()
+	agentPool, err := a.agentPoolsClient.Get(ctx, a.resourceGroupName, a.managedClusterName, a.agentPoolName)
```
piyush-nimbalkar: Ideally, we should not be storing the managedClusterName and agentPoolName on the azureOps instance. Every request that comes to this instance of azureOps may not be for the same managedCluster or agentPool.
The API is to inspect the InstanceGroup and we should find that out here based on the given VM id.

azure/azure.go
```diff
@@ -46,11 +59,101 @@ var (
...
+	agentPoolName      string
+	disksClient        *compute.DisksClient
+	vmsClient          vmsClient
+	snapshotsClient    *compute.SnapshotsClient
+	agentPoolsClient   *containerservice.AgentPoolsClient
+}
+
+// NewAzureClient initializes cloudops driver for azure based on environment
+// variables or based on instance metadata info available inside Azure VM
+func NewAzureClient() (cloudops.Ops, error) {
```
piyush-nimbalkar: NewAzureClient does not tell what is different in this method than the other two - NewClient and NewEnvClient.
May be call it NewClientFromMetadata?

azure/azure.go
vinayakshnd: @piyush-nimbalkar yah, update is working properly here just with the properties to be updated( size in this case.)

azure/azure.go
vinayakshnd: If timeout is zero then we want `SetInstanceGroupSize()` API to wait infinitely. If user wants API to be timebound then he needs to set non-zero timeout. 
If timeout is zero then no deadline is set to context `ctx` and WaitForCompletionRef() will not terminate on deadline.

azure/azure.go
vinayakshnd: what if we return count 0 with error if `*agentPool.Count` is nil? API consumer can check for err before consuming valume

azure/azure.go
vinayakshnd: Inspecting Instance Group purely using VM id is not possible since cloud sdk require other info like managedCluster. If we don't want to store that info in azureOps , then we may need to change API interface definition and it's other implementations

azure/azure.go
piyush-nimbalkar: Just check the nil error as discussed and return any default value (0/1) as per your requirement in the consuming code.

azure/azure.go
piyush-nimbalkar: Yes I understood that, just change non-zero -> zero in the comment.

azure/azure.go
piyush-nimbalkar: As discussed offline, let's leave it in the ops object for now as we have the instance stored as well. Eventually we should try to detach instance specific information from cloudops, that may require API changes.

azure/azure.go
vinayakshnd: @piyush-nimbalkar currently zone field is not used in any other API other than in test here:
https://github.com/libopenstorage/cloudops/blob/master/test/cloudops.go#L115

For the above test, if empty/nil availability zone array comes, I still need `len(groupInfo.Zones)` to be 1 to calculate the expected size of cluster.

azure/azure.go
```diff
@@ -67,14 +168,30 @@ func NewEnvClient() (cloudops.Ops, error) {
...
+	agentPoolName := os.Getenv(envAgentPoolName)
+
+	return NewClientWithPoolName(instance, scaleSetName, subscriptionID, resourceGroupName,
+		managedClusterName, agentPoolName)
 }
 
 // NewClient creates new client from specified parameters.
 func NewClient(
 	instance, scaleSetName, subscriptionID, resourceGroupName string,
+	managedClusterName string,
```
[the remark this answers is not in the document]
vinayakshnd: removed `managedClusterName string` parameter

azure/azure.go
vinayakshnd: added check for `nil` pointer here

azure/azure.go
vinayakshnd: Updated comment for both the cases.

azure/azure.go
vinayakshnd: Done

azure/azure.go
vinayakshnd: Done

azure/azure.go
vinayakshnd: Renamed function

piyush-nimbalkar:
Squash the commits before you merge.

PR #53, opened 2020-04-01
PWX-11697: Implement the Expand API

trierra:


gce/gce.go
```diff
@@ -624,27 +625,58 @@ func (s *gceOps) Expand(
...
+		Factor:   2,
+		Steps:    10,
+	}
+
+	checkForResize := func() (bool, error) {
+		newOp, err := s.computeService.ZoneOperations.Get(s.inst.project, s.inst.zone, fmt.Sprintf("%d", op.Id)).Do()
+		if err != nil {
+			return false, err
 		}
+		if newOp.Status == "DONE" {
```
adityadani, line 662: nit: set DONE as a constant

gce/gce.go
```diff
@@ -624,27 +625,58 @@ func (s *gceOps) Expand(
...
 	}
 
-	var disks []interface{}
-	for _, id := range diskNames {
-		if d, ok := allDisks[*id]; ok {
-			disks = append(disks, d)
-		} else {
-			return nil, fmt.Errorf("disk %s not found", *id)
+	// Taken from k8s.io/legacy-cloud-providers/aws
+	backoff := wait.Backoff{
```
adityadani, line 651: Lets use the same backoff as defined here: 
https://github.com/kubernetes/legacy-cloud-providers/blob/cebac2e3367faa71a39050bf5563fa7406006e76/gce/gce.go#L869

```
	backoff := wait.Backoff{
		// These values will add up to about a minute. See #56293 for background.
		Duration: time.Second,
		Factor:   1.4,
		Steps:    10,
	}
```

PR #57, opened 2020-05-08
Create azure clients using baseURI

piyush-nimbalkar:
- To support azure clouds like USGovernment, China, Germany clouds, azure sdk expects the clients to pass the management base URI when creating clients.
- Creating a new config object to be passed to NewClient as it is taking lot of args already.
- Changed NewClientFromMetadata to query only compute metadata

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -34,14 +34,13 @@ const (
 	envManagedClusterName = "AZURE_MANAGED_CLUSTER_NAME"
 	envAgentPoolName      = "AZURE_AGENT_POOL_NAME"
 	envUserAgent          = "AZURE_HTTP_USER_AGENT"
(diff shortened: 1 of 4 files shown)


azure/azure.go
```diff
@@ -1152,3 +1164,19 @@ func isExponentialError(err error) bool {
 	}
 	return false
 }
+
+func azureBaseURI(cloudEnvName string) (string, error) {
+	if value, ok := os.LookupEnv(auth.EnvironmentName); ok {
```
piyush-nimbalkar, line 1169: Ideally we don't need this as we get the azure cloud environment from the instance metadata. Keeping this in case the instance metadata (older versions did not have the azEnvironment metadata variable) does not have the field and user wants to explicitly override it.

PR #58, opened 2020-05-22
Keep disk

harsh-px:
This now creates vmdks with flag to not delete the disk on vm deletion.

Create code borrowed from: https://github.com/vmware/govmomi/blob/master/govc/disk/create.go#L82


vsphere/vsphere.go
```diff
@@ -732,3 +742,134 @@ func isExponentialError(err error) bool {
...
+
+func (c *common) ProcessOnce(fn func() error) (err error) {
+	c.process.Do(func() {
+		err = fn()
+	})
+	return err
+}
+
+// loopA runs before Sink() has been called.
+func (p *progressLogger) loopA() {
```
adityadani, line 805: why do we need two loops and a channel of channel ?
Can't we only use loopB and let Sink() return a channel p.sink where p.sink = make(chan progress.Report) ?

vsphere/vsphere.go
```diff
@@ -732,3 +742,134 @@ func isExponentialError(err error) bool {
...
+	var ok bool
+	var err error
+
+	for ok = true; ok; {
+		select {
+		case r, ok = <-ch:
+			if !ok {
+				break
+			}
+			err = r.Error()
```
adityadani, line 848: Don't we need to check r != nil here just like in the tick.C case ?
Also if we get an error shouldn't we exit out of this loop ?

harsh-px: removed the progress feature all together.

vsphere/vsphere.go
```diff
@@ -145,39 +149,45 @@ func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (inter
...
+	res, err := task.WaitForResult(ctx, logger)
+	logger.Wait()
 	if err != nil {
-		logrus.Errorf("Failed to get canonical vsphere disk path for: %s with "+
-			"volumeOptions: %+v on datastore: %s. err: %+v", diskPath, volumeOptions, datastore, err)
 		return nil, err
 	}
 
-	disk.DiskPath = canonicalVolumePath
+	myDisk := res.Result.(types.VStorageObject)
```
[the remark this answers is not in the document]
harsh-px: done for `res`

vsphere/vsphere.go
```diff
@@ -145,39 +149,45 @@ func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (inter
...
+	logrus.Infof(myDisk.Config.Id.Id)
```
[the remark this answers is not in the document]
harsh-px: done

vsphere/vsphere.go
harsh-px: removed the progress logger. It was optional and overly complex. Had inherited this from the govc sample code.

harsh-px:
Updated to PR to be backward compatible with < vSphere 6.7.3 versions

harsh-px:
rebased with ram's master changes and fixed conflicts

vsphere/vsphere.go
```diff
@@ -149,35 +152,95 @@ func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (inter
...
+		fileInfo, ok := myDisk.Config.Backing.(*types.BaseConfigInfoDiskFileBackingInfo)
+		if !ok {
+			return nil, fmt.Errorf("failed to get disk file path: %v", myDisk)
+		}
+		disk = diskmanagers.VirtualDisk{
+			DiskPath:      fileInfo.FilePath,
+			VolumeOptions: volumeOptions,
+		}
+	} else {
+		logrus.Warnf("Detected older vSphere version: %s. If VM gets deleted without detaching disk, disk "+
```
adityadani, line 212: Ideally this should be a warning/alert. Can we do a similar check in porx ?

vsphere/vsphere.go
```diff
@@ -149,35 +152,95 @@ func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (inter
...
 	volumeOptions.Datastore = datastore
 
-	diskBasePath := filepath.Clean(ds.Path(diskDirectory)) + "/"
-	err = ds.CreateDirectory(ctx, diskBasePath, false)
-	if err != nil && err != vclib.ErrFileAlreadyExist {
-		logrus.Errorf("Cannot create dir %#v. err %s", diskBasePath, err)
-		return nil, err
-	}
+	var (
+		trueVar = true
```
adityadani, line 156: nit: this could be renamed to keepDiskAfterDeleteVm

vsphere/vsphere.go
```diff
@@ -149,35 +152,95 @@ func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (inter
...
+		if err != nil {
+			return nil, err
+		}
 
-	disk.DiskPath = canonicalVolumePath
+		if res == nil {
+			return nil, fmt.Errorf("got empty result while creating vSphere disk")
+		}
+
+		myDisk := res.Result.(types.VStorageObject)
```
adityadani, line 200: nit: nil check

vsphere/vsphere.go
harsh-px: Removing it from here and will add as an error log and alert in the porx PR. Removing so that we don't have duplicate logs.

vsphere/vsphere.go
harsh-px: add check for res.Result too.

vsphere/vsphere.go
harsh-px: done

PR #65, opened 2020-07-10
For first class disks, use disk type provided by the user

harsh-px:
For vSphere 6.7 and above, we now create first class disks (FCD). In this mode, we are ignoring the user-provided disk type and defaulting to thin.


ram-infrac:
We were always creating thin type disk for FCD and now we support other disk type creation as well

harsh-px:
> We were always creating thin type disk for FCD and now we support other disk type creation as well

That's correct. Also thin type was not a good default as it can cause space management issues once disks start filling up. Eager zeroed thick is a better default.

PR #76, opened 2021-01-21
Add CSI provider support

trierra:
- CSI provider name
- CSI decision matrix
- CSI Storagemanager

### DIFF
--- a/specs/decisionmatrix/csi.yaml
+++ b/specs/decisionmatrix/csi.yaml
@@ -1,0 +1,10 @@
+rows:
+  - instance_type: "*"
+    instance_min_drives: 1
+    instance_max_drives: 12
+    region: "*"
+    min_size: 32
+    max_size: 8192
+    priority: 0
+    thin_provisioning: true
+    drive_type: "csi"
--- a/cloudops.go
+++ b/cloudops.go
@@ -24,6 +24,8 @@ const (
 	Vsphere = "vsphere"
 	// GCE provider
 	GCE = "gce"
+	// CSI provider
+	CSI = "csi"
 )
 
 // CloudResourceInfo provides metadata information on a cloud resource.
--- a/csi/storagemanager/csi.go
+++ b/csi/storagemanager/csi.go
@@ -1,0 +1,61 @@
+package storagemanager
+
+import (
+	"github.com/libopenstorage/cloudops"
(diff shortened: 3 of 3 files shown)


specs/decisionmatrix/csi.yaml
```diff
@@ -0,0 +1,10 @@
+rows:
+  - instance_type: "*"
+    instance_min_drives: 1
+    instance_max_drives: 12
+    region: "*"
+    min_size: 32
+    max_size: 8192
```
harsh-px, line 7: we can double the max size to go upto 16 TiB. I just saw the aws and azure yaml files and both go upto 16.

trierra: aren't we going on max for vshpere?

adityadani:
The decision matrix yaml would depend on the CSI storage provider. You will need to pass the actual provider name here and then select the corresponding yaml file for it.

harsh-px:
> The decision matrix yaml would depend on the CSI storage provider. You will need to pass the actual provider name here and then select the corresponding yaml file for it.

Since we don't know the drive type, didn't see much value in having matrix rows. For CSI, we only have the storage class.

specs/decisionmatrix/csi.yaml
harsh-px: If aws and azure are going upto 16 TiB, no reason to stop at 8 for vsphere. So vsphere should bump up too. But that can be for another day.

adityadani:
But for providers like vsphere, Pure FA where this CSI driver would be used, we can derive the drive type from the storage class and then based on that choose our distrbution matrix. The storage class parameters and values are always well defined. So instead of passing the drive type as csi, we will pass in the actual drive types based on the storage class. 

We might need another layer for this translation, but if we just have one row then it defeats the purpose of this decision matrix. For vsphere specifically we already have a decision matrix, and this CSI implementation should just use that.

trierra:
it's not required to provide a drive type in storage class. The minimum requirement is provisioner name. So there is a case for "generic" implementation. In CSI layer we have the defaulResolver for cases, if we  don't know a provider yet. Let's keep this one as a "defaultMatrix". In the future we will add per-provider matrix

adityadani (approved):
Discussed offline on this topic with Oksana.

PR #81, opened 2021-04-13
[PTX-2162] Specify OSType while setting instance group size for Azure

vinayakshnd:
This PR mentions the OSType of AKS node group while updating it. 

Without this fix updating of instance group size on AKS fails with below error:
```
failed to set size of node pool nodepool1. Error: Failed to set size for node group [nodepool1] of managed cluster [torpedo-tp-aks].Error:[containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="SystemPoolMustBeLinux" Message="System node pool must be linux. Nodepool name: nodepool1."] 
```
Closes #


vinayakshnd:
Jenkins run that used this change: https://jenkins.portworx.dev/job/Torpedo/job/tp-aks-asg/56/console

adityadani (approved):

PR #84, opened 2021-05-20
Updates Pure matrix to allow up to 32 drives per instance, and changes the Pure drive type to pure-block

Pure-AdamuKaapan:
Currently the Pure decision matrix only allows up to 12 drives, but we support up to (at least) 255 volumes connected at a time. This updates the matrix to reflect this.

### DIFF
--- a/specs/decisionmatrix/pure.yaml
+++ b/specs/decisionmatrix/pure.yaml
@@ -1,7 +1,7 @@
 rows:
   - instance_type: "*"
     instance_min_drives: 1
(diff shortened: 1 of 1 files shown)


harsh-px:
255 as a limit is too high. Let's keep a soft limit of 32.

255 makes sense for user volumes. This is uses for PX backing drives and we don't want a single PX node to have 255 backing drives.

harsh-px (approved):

PR #86, opened 2021-06-02
PWX-20131: Increase soft-limit on disk size to 32TiB on Azure

nrevanna:
Azure has new disk sizes/types since the last time we updated our yaml.
   This patch those new disk sizes and types

Closes #

### DIFF
--- a/azure/storagemanager/testspecs/azure-storage-decision-matrix.yaml
+++ b/azure/storagemanager/testspecs/azure-storage-decision-matrix.yaml
@@ -83,7 +83,29 @@ rows:
           instance_min_drives: 1
           region: "*"
           min_size: 8192
-          max_size: 8192
+          max_size: 16384
+          priority: 2
+          thin_provisioning: false
+          drive_type: "Premium_LRS"
+        - min_iops: 18000
+          max_iops: 18000
+          instance_type: "*"
+          instance_max_drives: 8
+          instance_min_drives: 1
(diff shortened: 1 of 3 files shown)


azure/storagemanager/testspecs/azure-storage-decision-matrix.yaml
```diff
@@ -87,6 +87,28 @@ rows:
           priority: 2
           thin_provisioning: false
           drive_type: "Premium_LRS"
+        - min_iops: 18000
```
adityadani, line 90: The previous row should now be
```
        - min_iops: 16000
          max_iops: 16000
          instance_type: "*"
          instance_max_drives: 8
          instance_min_drives: 1
          region: "*"
          min_size: 8192
          max_size: 16384
```
and so on... 

the last row will now be
```
        - min_iops: 20000
          max_iops: 20000
          instance_type: "*"
          instance_max_drives: 8
          instance_min_drives: 1
          region: "*"
          min_size: 32768
          max_size: 32768
          priority: 2
          thin_provisioning: false
          drive_type: "Premium_LRS"
```

nrevanna: @adityadani Do you want me to follow that pattern for other disk types too? They haven't been following it previously.

harsh-px (approved):

PR #87, opened 2021-08-04
PWX-17627: Adding IMDSv2 support

nrevanna:
Adding IMDSv2 support. Removing custom metadata function in favor of
      aws api available in the sdk.
      More info: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

Closes #PWX-17627

Testing: Create a aws cluster with nodes allowing IMDSv2 only. Was able to run px on such a cluster.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -2,8 +2,6 @@ package aws
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


adityadani (approved):
Changes look good. Do we need to update the vendor'ed aws library?

nrevanna:
> Changes look good. Do we need to update the vendor'ed aws library?
@adityadani 
I tested with what was already vendor'ed and it worked so I don't think we need to. However, minio package needed to be updated. There were some govendor/gomodules issue with the update. So, I cherry-picked the v2 only changes.

PR #89, opened 2021-12-08
PWX-21720 Dereference pointers for logs

dahuang-purestorage:
Signed-off-by: dahuang <dahuang@purestorage.com>

The logs were printing pointers, which doesn't mean much for human readers. This PR dereference the pointers for better logging.

Closes #

### DIFF
--- a/backoff/exponential.go
+++ b/backoff/exponential.go
@@ -1,7 +1,6 @@
 package backoff
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


nrevanna (approved):

backoff/exponential.go
```diff
@@ -377,7 +378,13 @@ func (e *exponentialBackoff) Enumerate(volumeIds []*string,
 	)
 	conditionFn := func() (bool, error) {
 		enumerateResponse, origErr = e.cloudOps.Enumerate(volumeIds, labels, setIdentifier)
-		msg := fmt.Sprintf("Failed to enumerate drives (%v).", volumeIds)
+		volumeIdsStr, err := volumeIdsStringDereference(volumeIds)
+		var msg string
+		if err != nil {
+			msg = err.Error()
```
nrevanna, line 384: I don't think you should change the message here. Instead, you can print the err here but you should retain the message. In line 388 we are responding to the origErr and not err. So if origErr is not nil we need a message that says what are the drives that couldn't enumerate. If one of the drives happened to be nil, we should just print that and let the person debugging take a look at it.
Additionally, in line 401, instead of returning, you should continue and make that volume name to be "". I don't expect to have a nil volume id since we would most likely crash/catch that in line 380. Even if we somehow have a nil volume id, let's not stop printing the rest of them.

dahuang-purestorage: good point. changed the check to continue when it sees a nil. and you are right, we should expect the cloudprovider to handle the nil case for us.

nrevanna (approved):

PR #91, opened 2022-01-05
PWX-21358: vsphere best effort even disk distribution

nrevanna:
There are two commits to this PR
1. Look for changes to the file `vsphere/vsphere.go`. The one with title "PWX-21358: vsphere best effort even disk distribution"
2. Vendor update. Title: "PWX-21358: Part 2 of updating to latest of portworx/sched-ops"

PWX-21358
Closes #

https://github.com/portworx/porx/pull/8577 has testing details. Copying screenshots was a challenge. Hence linking it.


vsphere/vsphere.go
```diff
@@ -729,7 +742,9 @@ func (ops *vsphereOps) getDatastoreToUseInStoragePod(
 	}
 
 	spec.DeviceChange = deviceChange
+	ops.dsLock.LockWithKey(ops.cfg.VMUUID, storagePod.Name())
 	recommendedDatastore, err := recommendDatastore(ctx, vmObj, storagePod, spec)
+	ops.dsLock.UnlockWithKey(storagePod.Name())
```
trierra, line 747: didn't quite get the lock with owner logic. Is the unlock with key sufficient? I assume it is, just trying to understand the way it works

vsphere/vsphere.go
```diff
@@ -796,6 +811,8 @@ func recommendDatastore(
 		PodSelectionSpec: podSelectionSpec,
 		ConfigSpec:       spec,
 		ResourcePool:     &resourcePoolRef,
+		// lock the recommended resources for us for the next 90 seconds
+		ResourceLeaseDurationSec: 90,
```
trierra, line 815: it seems like this var is not used anywhere

trierra (approved):
lgtm

vsphere/vsphere.go
nrevanna: Owner is for debugging purposes. Key defines a single lock. Since the key is pod name, if there are more than one pod at some point, there will be different locks for each pod.

vsphere/vsphere.go
```diff
@@ -77,12 +83,19 @@ func NewClient(cfg *VSphereConfig) (cloudops.Ops, error) {
 	logrus.Debugf("  Datacenter: %s", vmObj.Datacenter.Name())
 	logrus.Debugf("  VMUUID: %s", cfg.VMUUID)
 
+	configMap, err := configmap.New(vSphereDataStoreLock, nil, 0, 420, 5*time.Second, 10*time.Second)
```
adityadani, line 86: nit: would be good to define these as constants

vsphere/vsphere.go
adityadani: What happens if there is a datastore cluster with a single datastore? Will the other nodes have to wait for 90s until they get a recommendation for that datastore? 
Also why 90?

adityadani (commented):
looks good.. just a few questions.

vsphere/vsphere.go
nrevanna: Will send out a new PR for them.

nrevanna:
@adityadani 
90 seconds is the time for which vsphere will assume the datastore to have higher space utilization and not necessarily locking the datastore from being used by someone. 
e.g. Let's start with the datastore being completely free. And if we ask for a recommendation for a 200G vmdk, vsphere will assume that the datastore that it recommended will have a space utilization of 200G, even though the disk is not really placed on that datastore. With 200G being the new assumed size, it will run it's algorithm for any new recommendations that were asked. It will not stop itself from recommending the same datastore if that is the only one available. If the vmdk is not placed on the datastore within 90 seconds, it will subtract that 200G from that datastore and will continue with it's recommendation.

I saw that the CreateDisk was taking anywhere from 60 seconds - 90 seconds in my testing hence I chose the higher number.

PR #96, opened 2022-04-04
PWX-22761 Added support for IBM cloud

vinayakshnd:

Closes #


ibm/ibm_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if ibm.IsDevMode() {
+		drivers := make(map[string]cloudops.Ops)
+		diskTemplates := make(map[string]map[string]interface{})
+
+		d, disks := initIBM(t)
+		drivers[d.Name()] = d
+		diskTemplates[d.Name()] = disks
+		test.RunTest(drivers, diskTemplates, sizeCheck, t)
+	} else {
+		fmt.Printf("skipping GCE tests as environment is not set...\n")
```
adityadani, line 51: nit: skipping IBM tests

ibm/ibm.go
```diff
@@ -0,0 +1,209 @@
...
+
+	instanceInfo := &cloudops.InstanceInfo{
+		CloudResourceInfo: cloudops.CloudResourceInfo{
+			Name: workerDetails.ID,
+			Labels: map[string]string{
+				labelWorkerPoolName: workerDetails.PoolName,
+				labelWorkerPoolID:   workerDetails.PoolID,
+			},
+			Zone:   workerDetails.Location,
+			Region: workerDetails.Location,
```
adityadani, line 108: should this be workerDetails.Region ?

adityadani (approved):
lgtm

ibm/ibm.go
vinayakshnd: IBM SDK returns location as the only thing for worker details. Hence, populationg same for both zone and region for now.  As of now, there is no dependacy on these fields in test code.

ibm/ibm_test.go
vinayakshnd: done

PR #97, opened 2022-04-20
[PWX-22761]  Made IBM cloudops implementation self aware

vinayakshnd:
1. Made IBM cloudops implementation self aware of, on which node and cluster it is running.
2. Made `SetInstanceGroupSize()` API synchronus such that it waits for operation to complete before returning

Closes #


ram-infrac (approved):
lgtm

PR #98, opened 2022-06-16
IBM: Add a decision matrix file for IBM.

adityadani:
- IBM has 3 tiers of storage
- 3iops-tier: 3 IOPS per GB
- 5iops-tier: 5 IOPS per GB
- 10iops-tier: 10 IOPS per GB
Each of this tier has its own capacity limits and the decision matrix takes that into
account.

Auto-generate this file through go code.

Closes #


trierra (approved):

ibm/storagemanager/spec/generator.go
```diff
@@ -0,0 +1,98 @@
...
+		row.MaxSize = uint64(row.MaxIOPS / iopsMultiplier)
+		rows = append(rows, row)
+	}
+	return rows
+}
+
+func getCommonRow(priority int) cloudops.StorageDecisionMatrixRow {
+	return cloudops.StorageDecisionMatrixRow{
+		InstanceType:      "*",
+		InstanceMaxDrives: 8,
```
nrevanna, line 93: Is this number doesn't matter right now right? If it does, is this accurate? Couldn't find it in the document.

nrevanna (approved):
LGTM except for one comment.

ibm/storagemanager/spec/generator.go
adityadani: Which document are you referring to? We do have a runtime option for IBM provider which allows limiting to only 3 drives per node due to IBM limitations. But since this matrix is not used at install time today this max limit is not getting enforced.

PR #100, opened 2022-07-25
Added CloudOps implementation for Oracle Cloud

vinayakshnd:

Closes #


oracle/oracle.go
```diff
@@ -0,0 +1,235 @@
...
+	"os"
+
+	"github.com/libopenstorage/cloudops"
+	"github.com/oracle/oci-go-sdk/v65/common"
+	"github.com/oracle/oci-go-sdk/v65/containerengine"
+	"github.com/oracle/oci-go-sdk/v65/core"
+)
+
+const (
+	metadataAPIEndpoint           = "http://169.254.169.254/opc/v2/instance/"
```
nrevanna, line 18: I see that you are using the Metadata version 2 directly. We might want to evaluate a few things first.
Not all images will support v2: https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm#upgrading-v2__supported-images
We might also want to support v1 if we end up with one of the unsupported images.

vinayakshnd: Difference between v2 and v1 is, v2 requires additional request header called `Authorization: Bearer Oracle` and a different endpoint. JSON Response returned by both v1 and v2 is same. Execution of tests passed when I changed the `metadataAPIEndpoint ` to  `"http://169.254.169.254/opc/v2/instance/"`

May be lets perform metadata lookup to v1 endpoint first and if it fails then use v2.

vinayakshnd: Sure.

vinayakshnd: Using HTTP 404 NOT FOUND response to detect whether endpoint is enabled is more easier than to use GetInstance() API since 2nd approach requires compute client which again depends on certain things like region and tenancy ID that we are planning to get from metadata service itself than from user input.

oracle/oracle.go
```diff
@@ -12,25 +12,31 @@ import (
...
-	metadataAvailabilityDomainKey = "availabilityDomain"
+	v1MetadataAPIEndpoint = "http://169.254.169.254/opc/v1/instance/"
+	v2MetadataAPIEndpoint = "http://169.254.169.254/opc/v2/instance/"
+	metadataInstanceIDkey = "id"
+	// MetadataRegionKey key name for region in metadata JSON
+	// returned by IMDS service
+	MetadataRegionKey = "canonicalRegionName"
+	// MetadataAvailabilityDomainKey key name for availability domain
+	// in metadata JSON returned by IMDS service
+	MetadataAvailabilityDomainKey = "availabilityDomain"
```
nrevanna, line 26: Let's stick with this for now. May be during display we can think of getting rid of the tenancy name. Can you file a ticket to revisit it later. We'll have to see in our commands output and make sure it makes sense.

oracle/oracle.go
```diff
@@ -49,189 +55,174 @@ type oracleOps struct {
...
-		compartmentID:      compartmentID,
-		tenancyID:          tenancyID,
-		poolID:             poolID,
-	}, nil
+	return oracleOps, nil
 }
 
-func getInfoFromEnv() (string, string, string, string, string, string, error) {
-	instance, err := cloudops.GetEnvValueStrict(envInstanceID)
+func getInfoFromEnv(oracleOps *oracleOps) error {
```
nrevanna, line 87: Much cleaner. Thanks for refactoring.

oracle/oracle.go
```diff
@@ -49,189 +55,174 @@ type oracleOps struct {
...
+	httpHeaders["Authorization"] = "Bearer Oracle"
+	var httpStatusCode int
+	var err error
+	var metadata map[string]interface{}
+	metadata, httpStatusCode, err = GetRequest(v2MetadataAPIEndpoint, httpHeaders)
+	if err != nil {
+		return nil, err
+	}
+
+	if httpStatusCode == http.StatusNotFound {
```
nrevanna, line 175: Shouldn't we try for anything that is not StatusOK?

oracle/oracle.go
```diff
@@ -49,189 +55,174 @@ type oracleOps struct {
...
+		// Authorization header is safely ignored by v1 endpoint
+		fmt.Printf("Trying v1 endpoint as got %d http response\n", httpStatusCode)
+		metadata, httpStatusCode, err = GetRequest(v1MetadataAPIEndpoint, httpHeaders)
```
nrevanna, line 178: we don't need the httpHeaders for v1 right? Need to send ""?

oracle/oracle.go
vinayakshnd: Filed https://portworx.atlassian.net/browse/PWX-26183 for this.

oracle/oracle.go
vinayakshnd: Even if we pass httpHeaders for v1, it simply ignores them.

oracle/oracle.go
vinayakshnd: Done

oracle/oracle.go
```diff
@@ -0,0 +1,377 @@
...
+	var metadata map[string]interface{}
+	metadata, httpStatusCode, err = getRequest(v2MetadataAPIEndpoint, httpHeaders)
+	if err != nil {
+		return nil, err
+	}
+
+	if httpStatusCode != http.StatusOK {
+		// Authorization header is safely ignored by v1 endpoint
+		logrus.Warnf("Trying %s endpoint as got %d http response from %s\n",
+			v2MetadataAPIEndpoint, httpStatusCode, v1MetadataAPIEndpoint)
```
nrevanna, line 181: I think you need to swap v2MetadataAPIEndpoint and v1MetadataAPIEndpoint

oracle/oracle.go
```diff
@@ -0,0 +1,377 @@
...
+		metadata, httpStatusCode, err = getRequest(v1MetadataAPIEndpoint, httpHeaders)
```
nrevanna, line 182: I understand that it ignores the header. I don't get why we want to pass it though. Can't we just pass a blank map here and remove line 179?

oracle/oracle.go
```diff
@@ -0,0 +1,377 @@
...
+	if volumeAttachment.GetLifecycleState() != core.VolumeAttachmentLifecycleStateAttached {
+		return "", cloudops.NewStorageError(cloudops.ErrVolInval,
+			fmt.Sprintf("Invalid state %q, volume is not attached",
+				volumeAttachment.GetLifecycleState()), "")
+	}
+	if volumeAttachment.GetDevice() == nil {
+		return "", cloudops.NewStorageError(cloudops.ErrVolInval,
+			"Unable to determine volume attachment path", "")
+	}
+	// TODO: check do we really need to get actual device path.
```
nrevanna, line 352: Can we remove this comment since we agreed on using consistent paths?

oracle/oracle.go
```diff
@@ -0,0 +1,377 @@
...
+	oracleOps.compute, err = core.NewComputeClientWithConfigurationProvider(configProvider)
+	if err != nil {
+		return nil, err
+	}
+	oracleOps.containerEngine, err = containerengine.NewContainerEngineClientWithConfigurationProvider(configProvider)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: wrap around exponentialBackoffOps
```
nrevanna, line 85: Do we have a ticket tracking this?

vinayakshnd: Filed https://portworx.atlassian.net/browse/PWX-26214

oracle/oracle.go
vinayakshnd: Done

oracle/oracle.go
vinayakshnd: Done

oracle/oracle.go
vinayakshnd: Done

oracle/oracle.go
vinayakshnd: Passing empty header map now.

nrevanna (approved):
LGTM

PR #101, opened 2022-07-28
PWX-22292,pt1: Azure alternative Auth

zoxpx:
* adds ability for caller to specify external Authorizer for Azure API clients

Signed-off-by: Zoran Rajic <zrajic@purestorage.com>

The Azure's `NewCllient(config)` currently always generates the Authorizer from the environment variables.
* this PR is adding the ability to specify different client/secret/tenant AuthN via the config passed to `NewClient()`

Closes #  PWX-22292  (part1)

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -83,6 +83,7 @@ type Config struct {
 	ManagedClusterName string
 	AgentPoolName      string
 	UserAgent          string
+	Authorizer         autorest.Authorizer
 }
 
 // NewClientFromMetadata initializes cloudops driver for azure based on environment
@@ -206,9 +207,16 @@ func NewEnvClient() (cloudops.Ops, error) {
 
 // NewClient creates new client from specified config.
 func NewClient(config Config) (cloudops.Ops, error) {
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
-	if err != nil {
-		return nil, err
+	authorizer := config.Authorizer
+	if authorizer == nil {
+		logrus.Info("azure.NewClient authorizer generated from Env")
+		var err error
+		authorizer, err = auth.NewAuthorizerFromEnvironment()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		logrus.Info("azure.NewClient authorizer passed from caller")
 	}
 
 	baseURI, err := azureBaseURI(config.CloudEnvironment)
@@ -346,7 +354,7 @@ func (a *azureOps) SetInstanceGroupSize(instanceGroupID string,
 
 	instanceGroupSize := int32(count)
 	agentPoolProperties := containerservice.ManagedClusterAgentPoolProfileProperties{
-		Count: &instanceGroupSize,
+		Count:  &instanceGroupSize,
 		OsType: containerservice.Linux,
 	}
 
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -83,7 +83,12 @@ type Config struct {
 	ManagedClusterName string
 	AgentPoolName      string
 	UserAgent          string
-	Authorizer         autorest.Authorizer
+	// Auth is optional - if not provided, we'll use generic `auth.NewAuthorizerFromEnvironment`
+	Auth struct {
+		ClientID     string
+		ClientSecret string
+		TenantID     string
+	}
 }
 
 // NewClientFromMetadata initializes cloudops driver for azure based on environment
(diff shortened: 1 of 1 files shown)


azure/azure.go
```diff
@@ -206,9 +207,16 @@ func NewEnvClient() (cloudops.Ops, error) {
 
 // NewClient creates new client from specified config.
 func NewClient(config Config) (cloudops.Ops, error) {
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
-	if err != nil {
-		return nil, err
+	authorizer := config.Authorizer
```
adityadani, line 210: Can we create the authorizer here in the cloudsops layer itself, rather than asking the clients of cloudops to create an authorizer? In this way any user of cloudops (potentially operator) can leverage this new authorizer

zoxpx: OK -- this would also cut down on imported packages required by the caller

zoxpx: @adityadani -- your feedback addressed in updated [ed8c357](https://github.com/libopenstorage/cloudops/pull/101/commits/ed8c357446b7c95c6584bb260b1f5438a8f364ee)
* can pass string-values via config, to create a different Authorizer

adityadani (approved):
looks good. QQ: How does the client of cloudops get these clientIDs, ClientSecrets in this new model ?

zoxpx:
> looks good. QQ: How does the client of cloudops get these clientIDs, ClientSecrets in this new model ?

Oh.. that'll be in the "part 2" of the fix   (in `porx` project)

zoxpx:
Thanks for the review Aditya -- merging the PR into `master`-branch

azure/azure.go
```diff
@@ -206,9 +212,14 @@ func NewEnvClient() (cloudops.Ops, error) {
 
 // NewClient creates new client from specified config.
 func NewClient(config Config) (cloudops.Ops, error) {
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
-	if err != nil {
-		return nil, err
+	var authorizer autorest.Authorizer
+	var err error
+	if a := config.Auth; a.ClientID != "" && a.ClientSecret != "" {
```
piyush-nimbalkar, line 217: When authorizing using Managed identity, Azure only needs ClientID. So, let's just check for ClientID and not ClientSecret.

zoxpx: If one resets the password using `az ad sp credential reset --id $PRINCIPAL_UUID`, the `ClientID` and `TenantID` remain the same, but the `ClientSecret` changes.

Furthermore, the PX Azure config uses all 3 params -- i.e. https://install.portworx.com/2.11?mc=false&kbver=1.24.1&b=true&kd=type%3DPremium_LRS%2Csize%3D150&s=%22type%3DPremium_LRS%2Csize%3D150%22&c=px-cluster-XXX&aks=true&stork=false&csi=false&tel=false&st=k8s
```
      containers:
        - name: portworx
          image: portworx/oci-monitor:2.11.1
          env:
            - name: PX_NAMESPACE
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
            - name: AZURE_CLIENT_SECRET
              valueFrom:
                secretKeyRef:
                  name: px-azure
                  key: AZURE_CLIENT_SECRET
            - name: AZURE_CLIENT_ID
              valueFrom:
                secretKeyRef:
                  name: px-azure
                  key: AZURE_CLIENT_ID
            - name: AZURE_TENANT_ID
              valueFrom:
                secretKeyRef:
                  name: px-azure
                  key: AZURE_TENANT_ID
...
```

piyush-nimbalkar: @zoxpx that is the default way that most of our customers authorize. Managed identity in Azure is similar (not same) to instance permissions in AWS. We have some customers who use that to authorize instead of the standard client credentials way of auth.

For managed identity, they only need to pass the client_id and nothing else. Of course, they update the above default spec created by spec gen when they use this way of authenticating. Unfortunately, we don't document all different ways of authenticating. There are 4 ways, and client creds is just one way out of it.

zoxpx: @piyush-nimbalkar  Yes -- I have seen a few different ways to authenticate in the `auth.NewAuthorizerFromEnvironment()` (including the public/secret certs).  Still, how many way do we officially support via our YAML-configs, and the spec-generator, especially since we didn't document them?

I'm curious - if customer is using a managed identity, and needs to pass only the `$AZURE_CLIENT_ID` -- do they still need to edit the YAML-spec, and _remove_  the `$AZURE_CLIENT_SECRET` and `$AZURE_TENANT_ID` secrets-variables?  If they don't need to remove them, then our custom method is going to work.

piyush-nimbalkar: Yes, those customers edit the yaml spec to only pass the AZURE_CLIENT_ID. Rather they even create the secret with just client id because there is no client secret to pass in case of managed identity. That is why I said the above check of client secret won't work. 
If we just check the client_id, your change would work for most customer cases.

zoxpx:
Had a discussion with @piyush-nimbalkar -- will be rolling back this fix

PR #105, opened 2022-08-16
PWX-26307: Use default gce account when device level account not prov…

pp511:
…ided for encryption

Closes #PWX-26307 [https://portworx.atlassian.net/browse/PWX-26307]

https://github.com/portworx/porx/pull/9762 has already been created to add support for disk encryption using customer managed key.

Customer should be able to provide Encrypt/Decrypt on their default gce account for disk encryption so that if device level kms account is not provided then 
1. if an account.json file has been updated in the pod via GOOGLE_APPLICATION_CREDENTIALS, that service account should be used as disk encryption kms account.
2. If deployement is using instance gce account, that service account should be used as disk encryption kms account.

### DIFF
--- a/gce/gce.go
+++ b/gce/gce.go
@@ -2,6 +2,7 @@ package gce
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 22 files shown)


adityadani (approved):
lgtm

PR #106, opened 2022-08-17
Fix Oracle IMDS metadata lookup for non-oracle instances

vinayakshnd:
On non-oracle instances, http client for IMDS endpoint returns nil response. This commit handles the nil response.

Closes #


nrevanna:
How is this getting called in non-oracle instances?

vinayakshnd:
@nrevanna on non-oracle instances, http client returns error object with nil resp https://github.com/libopenstorage/cloudops/blob/master/oracle/oracle.go#L142. In this PR, I will just skip resp.StatusCode if response is nil.

nrevanna (approved):

PR #107, opened 2022-08-18
PWX-26383: Only log warning when instance service account is not extr…

pp511:
…acted in gke.

To fix Porx travis test failure

Closes #PWX-26383

Following changes were done to add support of default account for disk encryption
https://github.com/libopenstorage/cloudops/commit/07ce5dd9b4ac7f3ecf608dfaafb396eec9d54d47#diff-85d719939b78e1f1935dc08be3b96f95eb802c1f7644748ed1256ce94a29c522

We expect this function to not return an error if we are invoking this in non gke context.
We have similar calls in the same function logging only a warning https://github.com/libopenstorage/cloudops/blob/master/gce/gce.go#:~:text=inst.clusterLocation%2C%20err,%7D

### DIFF
--- a/gce/gce.go
+++ b/gce/gce.go
@@ -1240,11 +1240,11 @@ func gceInfo(ctx context.Context, inst *instance) error {
 	if content["client_email"] != nil {
 		inst.serviceAccount = fmt.Sprintf("%s", content["client_email"])
 	} else {
-		serviceAccount, err := metadata.Email("")
+		inst.serviceAccount, err = metadata.Email("")
 		if err != nil {
-			return fmt.Errorf("unable to get gce instance service account")
+			// No need to error out for non-GKE compute instances
+			logrus.Warnf("unable to get gce instance service account")
 		}
-		inst.serviceAccount = serviceAccount
 	}
 	return nil
 }


pp511:
- TRAVIS_CONTROL_PLANE_UTS run on gce instance. 
- There is no email metadata set on the travis instance. 
- Service account associated with the  gce instance on travis doesn’t have required privileges to make cloud api calls
-  If I create a GKE cluster without specifying an account, it still does get associated with a default compute engine account in the project and that can be extracted with the .Email() function.  
- On Travis however, it complains “GCE metadata \“instance/service-accounts/default/email\” not defined”.
-  I believe it makes sense to open a new ticket for further exploration of  updating the Travis test such that it uses the correct service account. 

Not returning an error in case of missing instance email I believe still make sense since for all use-cases thus far we haven’t required access to the service account email and initialization shouldn't fail due to missing email metadata.

nrevanna:
That sounds fair. Can we ensure that wherever serviceAccount is used, it correctly handles serviceAccount being an empty string.

pp511:
This param is only being passed during disk encryption currently. If we do not provide a service account, gke will default to default compute engine account in the project for kms encryption/decryption. So empty service account use-case is handled.

nrevanna (approved):

PR #112, opened 2022-09-02
PWX-24702 : Add cloudOps API to upgrade OKE cluster

nikita-bhatia:
Jira ticket : https://portworx.atlassian.net/browse/PWX-24702

Changes done : 
> Implemented SetClusterVersion API - update kubernetes version of the Cluster
> Implemented SetInstanceGroupVersion API - update kubernetes version of nodepool and all existing nodes

Tests Done : 
> Verified Upgrade scenarion from v1.21.5 to v1.24.1 in steps and by skipping steps, for cluster , nodepool and all nodes.

### DIFF
--- a/oracle/oracle.go
+++ b/oracle/oracle.go
@@ -786,3 +786,92 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
 	}
 	return false
 }
+
+func (o *oracleOps) SetClusterVersion(version string, timeout time.Duration) error {
+	req := containerengine.UpdateClusterRequest{
+		ClusterId: &o.clusterID,
+		UpdateClusterDetails: containerengine.UpdateClusterDetails{
+			KubernetesVersion: &version,
+		},
+	}
+
+	resp, err := o.containerEngine.UpdateCluster(context.Background(), req)
+	if err != nil {
+		return err
+	}
+
+	return o.waitTillWorkStatusIsSucceeded(resp.OpcRequestId, resp.OpcWorkRequestId, timeout)
+}
+
+func (o *oracleOps) SetInstanceGroupVersion(instanceGroupName string, version string, timeout time.Duration) error {
+
+	//get nodepool ID from name
+	var instanceGroupId *string
+	nodePoolsReq := containerengine.ListNodePoolsRequest{CompartmentId: &o.compartmentID, Name: &instanceGroupName, ClusterId: &o.clusterID}
+	nodePools, err := o.containerEngine.ListNodePools(context.Background(), nodePoolsReq)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


oracle/oracle.go
```diff
@@ -786,3 +786,92 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
...
+
+	if err != nil {
+		return err
+	}
+
+	err = o.waitTillWorkStatusIsSucceeded(resp.OpcRequestId, resp.OpcWorkRequestId, timeout)
+	if err != nil {
+		return err
+	}
+
```
vinayakshnd, line 838: Put a comment, why we need to scale down to 0 and scale back again. Also, add referance link to Oracle doc. that mentions the limitation.

oracle/oracle.go
```diff
@@ -786,3 +786,92 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
...
+	//get existing total node count of instanceGroup
```
vinayakshnd, line 839: Put all the steps of scaling down to 0 and scaling back to original size, in separate private helper function and just call that here.

oracle/oracle.go
```diff
@@ -786,3 +786,92 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
...
+		nodePoolPlacementConfigDetails[i].AvailabilityDomain = placementConfigs.AvailabilityDomain
+		nodePoolPlacementConfigDetails[i].SubnetId = placementConfigs.SubnetId
+	}
+	//delete all nodes from existing node pool
+	if err := o.SetInstanceGroupSize(instanceGroupName, 0, timeout); err != nil {
+		return err
+	}
+
+	//create same number of nodes again with updated version
+	req := containerengine.UpdateNodePoolRequest{
```
vinayakshnd, line 861: Can we use existing [SetInstanceGroupSize](https://github.com/libopenstorage/cloudops/blob/master/oracle/oracle.go#L476) function to set the size to original size?

vinayakshnd (changes requested):

oracle/oracle.go
nikita-bhatia: The SetInstanceGroupSize functions sets the cluster size per domain, we need functionality to create same number of instances as original. If number of instances are not in multiple of number of domains (edge case), the functionality might break. Hence not calling SetInstanceGroupSize here.

oracle/oracle.go
nikita-bhatia: Added

oracle/oracle.go
nikita-bhatia: Done

vinayakshnd (approved):
LGTM

PR #113, opened 2022-09-02
PWX-24576: Implement clouddrive Expand API in CloudOps

nikita-bhatia:
Jira ticket : https://portworx.atlassian.net/browse/PWX-24576

Changes done : 
Implemented Expand API for Oracle cloud

Tests done : 
Verified Expansion of disk with and without attachment to instance.

### DIFF
--- a/oracle/oracle.go
+++ b/oracle/oracle.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


oracle/oracle.go
```diff
@@ -786,3 +786,30 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
 	}
 	return false
 }
+
+func (o *oracleOps) Expand(volumeID string, newSizeInGiB uint64) (uint64, error) {
```
vinayakshnd, line 790: First check the current size of the disk. If requested `newSizeInGiB` is less than or equal to current size then return an error.
Take a look at how are we currently doing in AWS: https://github.com/libopenstorage/cloudops/blob/master/aws/aws.go#L937

vinayakshnd (changes requested):

oracle/oracle.go
```diff
@@ -786,3 +786,30 @@ func nodePoolContainsNode(s []containerengine.Node, e string) bool {
...
+	logrus.Println("Expand volume to size ", newSizeInGiB)
```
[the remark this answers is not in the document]
nikita-bhatia: Done

oracle/oracle.go
nikita-bhatia: Done

vinayakshnd (approved):
LGTM

PR #117, opened 2022-10-06
[PWX-24665] Modified `volume detached` condition for DevicePath() API for oracle provider.

vinayakshnd:
DevicePath() API should return `cloudops.ErrVolDetached` error in all of the below condition:

1. VolumeAttachment response have empty list of attachments. []
2. VolumeAttachment has no details of instance to which it is attached.
3. VolumeAttachment is in `DETACHING` or `DETACHED` state.

Closes #


pp511 (approved):
lgtm

PR #118, opened 2022-10-12
PWX-27046: Avoid panic when drive type is not specified.

pp511:

*vol.VolumeType results into nil pointer dereferencing. This occurs when drive type is not specified in the deviceSpec.
Also some indentation issues where auto fixed by the ide.

Closes # PWX-27046

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -786,7 +786,7 @@ func (s *awsOps) Create(
 	}
 	if vol.VolumeType == nil {
 		return nil, cloudops.NewStorageError(cloudops.ErrVolInval,
-			"Drive type not specified", "")
+			"Drive type not specified in the storage spec", "")
 	}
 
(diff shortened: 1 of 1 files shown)


aws/aws.go
```diff
@@ -784,6 +784,10 @@ func (s *awsOps) Create(
 		return nil, cloudops.NewStorageError(cloudops.ErrVolInval,
 			"Invalid volume template given", "")
 	}
+	if vol.VolumeType == nil {
+		return nil, cloudops.NewStorageError(cloudops.ErrVolInval,
+			"Drive type not specified", "")
+	}
```
nrevanna, line 790: nit: May be "Drive type not specified in the storage spec"?

nrevanna (approved):

aws/aws.go
pp511: Done.

PR #119, opened 2022-11-03
[pwx-27619] Add options for selected cloudops APIs

pure-jliao:

Closes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,9 @@ vendor-tidy:
 vendor:
 	go mod vendor
 
-lint:
-	(mkdir -p tools && cd tools && GO111MODULE=off && go get -v golang.org/x/lint/golint)
+$(GOPATH)/bin/golint:
+	GO111MODULE=off go get -u golang.org/x/lint/golint
+lint: $(GOPATH)/bin/golint
 	for file in $$(find . -name '*.go' | grep -v vendor | \
                                        grep -v '\.pb\.go' | \
                                        grep -v '\.pb\.gw\.go' | \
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -40,8 +40,9 @@ vendor-tidy:
 vendor:
 	go mod vendor
 
+$(GOPATH)/bin/golint:
+	GO111MODULE=off go get -u golang.org/x/lint/golint
 lint:
```
piyush-nimbalkar, line 45: Don't we need to include the `$(GOPATH)/bin/golint` task in this so that golint is downloaded when someone call `make lint`?

aws/aws.go
```diff
@@ -34,6 +34,9 @@ const (
 	awsDevicePrefixWithH = "/dev/hd"
 	awsDevicePrefixNvme  = "/dev/nvme"
 	contextTimeout       = 30 * time.Second
+
+	// DryRunOption is the key to tell if dry run the request
+	DryRunOption = "--dry-run"
```
piyush-nimbalkar, line 39: Instead of taking an AWS specific flag, let's define this at the cloudops interface level and a constant like `DryRun`. 
We should look at that flag in the aws provider and convert it to whatever aws wants. And ignore for the rest of the providers.

Makefile
pure-jliao: sorry seems it's missing. just because on my dev machine it would download a new lint version that will stuck, so I made this change according to operator.

piyush-nimbalkar (approved):

PR #120, opened 2022-11-03
[PWX-24574] Added GetInstance() API for oracle cloud provider

vinayakshnd:
Signed-off-by: Vinayak Shinde <vinayakshnd@gmail.com>

Closes #

DeleteInstance() API for oracle provider requires oracle compute instance ID. For torpedo, previously there was no cloudops API which would fetch instance details (and ID) from given instance name (Torpedo tracks instance by instance name). This API will help torpedo to fetch instance ID from given instance name.

### DIFF
--- a/oracle/oracle.go
+++ b/oracle/oracle.go
@@ -280,6 +280,26 @@ func (o *oracleOps) InspectInstance(instanceID string) (*cloudops.InstanceInfo, 
 	}, nil
 }
 
+func (o *oracleOps) GetInstance(displayName string) (interface{}, error) {
+	listInstanceReq := core.ListInstancesRequest{
+		DisplayName:   common.String(displayName),
+		CompartmentId: common.String(o.compartmentID),
+	}
+
+	listInstanceResp, err := o.compute.ListInstances(context.Background(), listInstanceReq)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


oracle/oracle.go
```diff
@@ -280,6 +280,22 @@ func (o *oracleOps) InspectInstance(instanceID string) (*cloudops.InstanceInfo,
...
+	}
+
+	listInstanceResp, err := o.compute.ListInstances(context.Background(), listInstanceReq)
+	if err != nil {
+		return nil, err
+	}
+	if len(listInstanceResp.Items) == 0 {
+		return nil, fmt.Errorf("no oracle instance found with display name: %s", displayName)
+	}
+	return listInstanceResp.Items[0], nil
```
nikita-bhatia, line 296: Can we have 2 instances with same displayname?

vinayakshnd: Currently, torpedo uses this API to fetch details of the instance created  by OKE. OKE ensures that all the worker nodes created, have unique display names. In future, if we require to use this API to get details of vanilla compute instances, then we can modify this array indexing.

nikita-bhatia (approved):
LGTM

PR #121, opened 2022-11-16
PWX-27769: Recognize all zone labels on vSphere and IBM

nrevanna:
On IBM and vSphere providers, not all the following zone labels are
      recognized.
      This patch adds support to recognize zones in the below order of
      labels:
        topology.portworx.io
        px/zone (for backward compatibility)
        topology.kubernetes.io
        failure-domain.beta.kubernetes.io

Signed-off-by: Naveen Revanna <nrevanna@purestorage.com>

Closes #

### DIFF
--- a/cloudops.go
+++ b/cloudops.go
@@ -2,13 +2,32 @@
 
 package cloudops
 
-import "time"
+import (
+	v1 "k8s.io/api/core/v1"
+	"time"
+)
 
 const (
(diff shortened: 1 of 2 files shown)


piyush-nimbalkar (commented):
I feel these labels are Portworx specific and should not be part of the cloudops library, but should reside in Portworx code itself.

PR #126, opened 2023-01-17
Added decision matrix row for max disk size for Oracle block volumes

vinayakshnd:
This PR enables user to expand Oracle block volumes beyond Max IOPS disk size.

Closes #


nrevanna (approved):
LGTM

PR #136, opened 2023-03-16
PWX-21520 : Add custom namespace during configmap creation 

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-21520

Changes done :
> Version of sched-ops is changed to latest commit hash to master. Because of that, other versions were also required to be updated and compatibility with Go 1.17 was not there anymore.

P.S : Need to update cloudops version in "porx" go.mod after this PR is merged.


sharma-tapas (approved):

store/store.go
```diff
@@ -15,9 +16,9 @@ const (
 // Params is the parameters to use for the Store object
 type Params struct {
 	// Kv is the bootstrap kvdb instance
-	Kv            kvdb.Kvdb
+	Kv kvdb.Kvdb
```
nrevanna, line 19: nit: The older formatting looked correct. Could you please keep it aligned like the previous version?

store/store.go
```diff
@@ -97,6 +98,7 @@ func GetStoreWithParams(
 	name string,
 	lockTryDuration time.Duration,
 	lockHoldTimeout time.Duration,
+	nameSpace string,
```
nrevanna, line 101: I feel it's not required to pass in namespace as parameter in all these functions. 
Why not just call `ns := os.Getenv("PX_NAMESPACE")` in newK8sStoreWithParams and pass that to configmap.New().

nrevanna (commented):
If you implement my comment, it'll be a 2 line patch. Why do we need to vendor so many files?

nrevanna (commented):
Actually, why don't we take a step back. Why not  call `ns := os.Getenv("PX_NAMESPACE")` from sched-ops itself. That way we don't need to pass in the namespace as a parameter to configmap.New?

nrevanna:
Had an offline discussion. We want to avoid PX_NAMESPACE usage in all our open-source repos. Nikita will work on removing that from cloud-ops also. 
Thanks Piyush.

PR #137, opened 2023-03-22
PWX-28882: Fix Nomad support issues for vSphere provider

nrevanna:
Configmap apis were used in vSphere provider which stopped px from
    coming up on Nomad. The following changes are made
        - Store interface moved to cloudops
        - DriveSetStore interface created with Store as composite. Contains
          interfaces with DriveSet datastructure.
        - GetLockTryDuration introduced
        - Renamed GetLockTimeout/SetLockTimeout to
          GetLockHoldTimeout/SetLockHoldTimeout.
        - Implemented IsKeyLocked() and LockWithKey() interfaces in kv_store
        - Removed noOpStore implementation. It was a stop gap solution for
          lack of implementation in kv_store

Closes #


store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+	return k8s.patchWithRetries(data)
+}
+
+func (k8s *k8sStore) PutKey(key string, value []byte) error {
+	data, err := k8s.cm.Get()
+	if err != nil {
+		return err
+	}
+
+	data[key] = string(value)
```
pureneelesh, line 173: can we safely assume that the value can always be converted to string without requiring any encoding e.g. base64? If yes, should we just change the interface to accept string instead of []byte to make that clear?

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, &StoreKeyDoesNotExist{
+			Key: key,
+		}
+	}
+
+	return []byte(value), nil
+}
+
+func (k8s *k8sStore) DeleteKey(key string) error {
+	sanitizedKey := GetSanitizedK8sName(key)
```
pureneelesh, line 194: I can't tell when we need to sanitize the key and when we don't. Also, not clear about when we need to lock. Please add comments to clarify. Also, functions could be named differently to distinguish if we are assuming that caller needs to have acquired a lock in some cases.

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	return returnKeys, nil
+}
+
+func (k8s *k8sStore) patchWithRetries(data map[string]string) error {
+	f := func() (bool, error) {
+		err := k8s.cm.Patch(data)
+
+		for _, retryErr := range etcdErrorsToRetry {
```
pureneelesh, line 240: do we actually see etcd errors from the k8s api server? I would have thought that the k8s API server will hide etcd errors from the clients.

store/store_kv.go
```diff
@@ -0,0 +1,138 @@
...
+// kvdb.
+func NewKVStoreWithParams(
+	kv kvdb.Kvdb,
+	name string,
+	lockTryDuration time.Duration,
+	lockHoldDuration time.Duration,
+) (Store, error) {
+	kstore := kvStore{}
+	if kv == nil {
+		return nil, ErrKvdbNotInitialized
```
pureneelesh, line 48: Name of the error seems misleading. It made me think that kvdb has not been setup yet. We could just use InvalidArgument or something like that?

vsphere/vsphere.go
```diff
@@ -83,10 +83,19 @@ func NewClient(cfg *VSphereConfig) (cloudops.Ops, error) {
 	logrus.Debugf("  Datacenter: %s", vmObj.Datacenter.Name())
 	logrus.Debugf("  VMUUID: %s", cfg.VMUUID)
 
-	configMap, err := configmap.New(vSphereDataStoreLock, nil, 0, 420, 5*time.Second, 10*time.Second)
-	if err != nil {
-		logrus.Errorf(err.Error())
-		return nil, err
+	var storeInstance store.Store
+	if storeParams != nil {
```
pureneelesh, line 87: what happens if storeParams is nil? Should we return an error?

vsphere/vsphere.go
```diff
@@ -755,9 +764,13 @@ func (ops *vsphereOps) getDatastoreToUseInStoragePod(
 	}
 
 	spec.DeviceChange = deviceChange
-	ops.dsLock.LockWithKey(ops.cfg.VMUUID, storagePod.Name())
+	storeLock, lockErr := ops.dsLock.LockWithKey(ops.cfg.VMUUID, storagePod.Name())
 	recommendedDatastore, err := recommendDatastore(ctx, vmObj, storagePod, spec)
-	ops.dsLock.UnlockWithKey(storagePod.Name())
+	// We are doing a best effort
+	if lockErr != nil {
+		ops.dsLock.Unlock(storeLock)
```
pureneelesh, line 771: should we call Unlock? I would not expect storeLock to have a valid value when lockErr is not nil. Perhaps just log a message?

store/store.go
```diff
@@ -0,0 +1,122 @@
...
+	var (
+		s          Store
+		err        error
+		withParams bool
+	)
+
+	withParams = (lockTimeout > 0) || (lockTryDuration > 0)
+	if internalKvdb && schedulerType == Kubernetes {
+		if withParams {
+			s, _, err = NewK8sStoreWithParams(name, lockTryDuration, lockTimeout)
```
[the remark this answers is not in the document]
nrevanna: I am doing that for NewKvStoreWithParams. 
configmap.New does that for NewK8sStoreWithParams

store/store.go
```diff
@@ -0,0 +1,122 @@
...
+	} else if internalKvdb && kv == nil {
+		return nil, fmt.Errorf("Bootstrap kvdb cannot be empty")
+	} else {
+		// Two cases:
+		// internal kvdb && kv is not nil
+		// external kvdb
+		if withParams {
+			s, err = NewKVStoreWithParams(kv, name, lockTryDuration, lockTimeout)
+		} else {
+			s, err = NewKVStore(kv)
```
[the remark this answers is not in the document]
nrevanna: I'll change to return error if the name is empty.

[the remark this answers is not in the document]
nrevanna: I misread. I didn't want to add the customization with name and default timeout params in order to avoid some duplication. Let me revisit.

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	return k8s.cm.Unlock()
+}
+
+func (k8s *k8sStore) IsKeyLocked(key string) (bool, string, error) {
+	return k8s.cm.IsKeyLocked(key)
+}
+
+func (k8s *k8sStore) CreateKey(key string, value []byte) error {
+	sanitizedKey := GetSanitizedK8sName(key)
```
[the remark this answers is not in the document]
nrevanna: GetSanitizedK8sName was added as part of a different request.
The only time it will result in same sanitized key for two different keys is when the two keys differ in a special character.
Key@ and Key$.
At the time of implementation we determined that special characters shouldn't be a differentiator. The idea of mapping each special character to some Letter was also pitched but wasn't perceived to be necessary.

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (k8s *k8sStore) PutKey(key string, value []byte) error {
+	data, err := k8s.cm.Get()
+	if err != nil {
+		return err
+	}
+
+	data[key] = string(value)
+	return k8s.patchWithRetries(data)
```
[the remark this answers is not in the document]
nrevanna: ref: PWX-27246
PutKey uses patchWithRetries which causes a race condition describe above.

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+	return k8s.cm.Unlock()
+}
+
+func (k8s *k8sStore) IsKeyLocked(key string) (bool, string, error) {
+	return k8s.cm.IsKeyLocked(key)
+}
+
+func (k8s *k8sStore) CreateKey(key string, value []byte) error {
+	sanitizedKey := GetSanitizedK8sName(key)
+	err := k8s.cm.LockWithKey(string(value), sanitizedKey)
```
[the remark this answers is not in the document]
nrevanna: We probably might have to do that at some point.
CreateKey is called with the hostname which could contain special characters which we had encountered and was fixed.
Today LockWithKey is mostly static names. Could possibly encounter the same issue in the future. I'll file a ticket to do that.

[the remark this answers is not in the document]
nrevanna: https://portworx.atlassian.net/browse/PWX-30053
I am looking to merge this today. I'll resolve all the comments I have. If I have time after that, I'll address this too. Just avoiding extra testing.

store/store_k8s.go
nrevanna: CreateKey and DeleteKey were only use in vSphere and vSphere hostname had special characters and that was the reason it was added. These are existing functions. If you have some specific suggestions I can do them. The larger effort I'd prefer to defer to a later time since I am on a time crunch.

store/store_k8s.go
nrevanna: I am not sure. May be @adityadani might have more context to this.

vsphere/vsphere.go
nrevanna: I'll probably do that and handle in the caller.

store/store_kv.go
nrevanna: This is from the existing code and I feel this it's technically correct to call ErrKvdbNotInitialized since it's an indication that the bootstrap kvdb is not initialized.

store/store.go
```diff
@@ -0,0 +1,122 @@
...
+	DeleteKey(key string) error
+	// EnumerateKey enumerates all keys in the store that begin with the given key
+	EnumerateKey(key string) ([]string, error)
+}
+
+// GetStoreWithParams returns instance for DriveSetStore
+// kv: bootstrap kvdb
+// schedulerType: node scheduler type e.g Kubernetes
+// internalKvdb: If the cluster is configured to have internal kvdb
+// name: Name for the configmap entry.
```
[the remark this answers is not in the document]
nrevanna: I corrected the comment and now calling to Store name. It's the name for which there will be an entry in the bootstrap kv.

store/store_k8s.go
nrevanna: It appears so. May be @adityadani can add more context to this.

store/store_kv.go
```diff
@@ -0,0 +1,140 @@
...
+	k                kvdb.Kvdb
+	keyName          string
+	lockTryDuration  time.Duration
+	lockHoldDuration time.Duration
+}
+
+// NewKVStore returns a Store implementation which is a wrapper over
+// kvdb.
+func NewKVStore(kv kvdb.Kvdb, name string) (Store, error) {
+	if len(name) == 0 {
```
adityadani, line 34: Lets fail the NewKVStore operation if name is empty and force the callers to set a key, so that we won't end up causing key conflicts

store/store_kv.go
```diff
@@ -0,0 +1,140 @@
...
+func (kv *kvStore) Unlock(storeLock *StoreLock) error {
+	kvp, ok := storeLock.internalLock.(*kvdb.KVPair)
+	if !ok {
+		return fmt.Errorf("Invalid StoreLock provided")
+	}
+	return kv.k.Unlock(kvp)
+}
+
+func (kv *kvStore) LockWithKey(owner, key string) (*StoreLock, error) {
+	fullPath := kv.keyName + "/" + key
```
adityadani, line 88: The previous code used to take a lock under `clouddrive/_lock/`. Should we follow the same suite to ensure locks are under a prefix? In this way we don't change the behavior for Lock() API

vsphere/vsphere.go
```diff
@@ -83,7 +88,13 @@ func NewClient(cfg *VSphereConfig) (cloudops.Ops, error) {
...
 	logrus.Debugf("  VMUUID: %s", cfg.VMUUID)
 
-	configMap, err := configmap.New(vSphereDataStoreLock, nil, 0, 420, 5*time.Second, 10*time.Second)
+	storeInstance, err := store.GetStoreWithParams(
+		storeParams.Kv,
+		storeParams.SchedulerType,
+		storeParams.InternalKvdb,
+		vSphereDataStoreLock,
+		420*time.Second,
+		150*time.Second)
```
adityadani, line 97: Should this be 180*time.Second instead to make it same as the default kvdb lock timeout of 3 minutes?

adityadani (approved):
lgtm after comments are addressed

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, fmt.Errorf("Bootstrap kvdb cannot be empty")
+	} else {
+		// Two cases:
+		// internal kvdb && kv is not nil
+		// external kvdb
+		if withParams {
+			s, err = NewKVStoreWithParams(kv, name, lockTryDuration, lockHoldTimeout)
+		} else {
+			s, err = NewKVStore(kv, name)
+		}
```
adityadani, line 124:
```suggestion
		// Two cases:
		// internal kvdb && kv is not nil
		// external kvdb
		if !internalKvdb {
		    if kvdb.Instance() == nil {// return kvdb is not initialized}
		    kv = kvdb.Instance()
		}
		if withParams {
			s, err = NewKVStoreWithParams(kv, name, lockTryDuration, lockHoldTimeout)
		} else {
			s, err = NewKVStore(kv, name)
		}
```

store/store_kv.go
adityadani: I think thats what you are trying to do. Can we rename this variable to `lockPrefix` ?

store/store_k8s.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	return k8sStore, cm, nil
+}
+
+// NewK8sStoreWithParams returns a Store implementation which uses
+// k8s configmaps to store data. ConfigMap properties can be customized.
+func NewK8sStoreWithParams(
+	name string,
+	lockTryDuration time.Duration,
+	lockTimeout time.Duration,
```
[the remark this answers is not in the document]
nrevanna: I thought it was not needed for any of the existing use cases. We can add when required. There might be some extra work for this in the kv_store side.

vsphere/vsphere.go
nrevanna: I'd probably reduce that to 100 seconds. The lock will kinda serialize node brinup, so I'd reduce it futher.

store/store_k8s.go
nrevanna: changed

store/store_k8s.go
nrevanna: https://portworx.atlassian.net/browse/PWX-30066

PR #140, opened 2023-03-27
PWX-29108 : Handle nil pointer dereference in cloudops

rge-pure:
Following issue was reported on gs-rel 2.13 on Azure [PWX-29069](https://portworx.atlassian.net/browse/PWX-29069). This was related to nil drive size being passed during drive create which was resulting into nil pointer dereference.  This was fixed for Azure.

We need to do a thorough check on all the supported cloud platforms supported in cloudops for all the params that can be passed as “nil” that can result into a panic and handle those.

**Which issue(s) this PR fixes**
[PWX-29018](https://portworx.atlassian.net/browse/PWX-29108)

Workflow
- Scan the entire repo for different cloudops
- Add nil check for potential nil parameters. 
- Report corresponding error logs.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -298,6 +298,10 @@ func (s *awsOps) InspectInstanceGroupForInstance(instanceID string) (*cloudops.I
 			}
 
 			group := result.AutoScalingGroups[0]
+			if group == nil {
+				return nil, fmt.Errorf("nil AutoScalingGroups group")
+			}
+
 			zones := make([]string, 0)
 			for _, z := range group.AvailabilityZones {
 				zones = append(zones, *z)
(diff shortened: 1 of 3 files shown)


pp511:
Discussed offline. Summarizing here:
Some params like *Encryption can be passed as nil. It won't cause panic but would just end up creating a non encrypted disk. Let's only check for mandatory params Eg: size. type could be a mandatory param in some clouds(not sure which) but is not in AWS as it defaults to gp2.

rge-pure:
Thanks Priyanshu for the advice.

aws/aws.go
```diff
@@ -796,7 +808,18 @@ func (s *awsOps) Create(
 		return nil, cloudops.NewStorageError(cloudops.ErrVolInval,
 			"Drive type not specified in the storage spec", "")
 	}
-
+	if vol.AvailabilityZone == nil {
```
rge-pure, line 811: According to CreateVolumeInput documentation
- AvailabilityZone is a required field
- The size of the volume, in GiBs. You must specify either a snapshot ID or a volume size.
- The snapshot from which to create the volume. You must specify either a snapshot ID or a volume size.

aws/aws.go
```diff
@@ -796,7 +808,18 @@ func (s *awsOps) Create(
 		return nil, cloudops.NewStorageError(cloudops.ErrVolInval,
```
rge-pure, line 808: It is not mandatory but being dereference below
`if *vol.VolumeType == opsworks.VolumeTypeIo1 || *vol.VolumeType == "gp3" {
`

azure/azure.go
```diff
@@ -371,14 +371,20 @@ func (a *azureOps) Create(
 	options map[string]string,
 ) (interface{}, error) {
 	d, ok := template.(*compute.Disk)
-	if !ok || d.DiskProperties == nil || d.DiskProperties.DiskSizeGB == nil {
+	if !ok || d.DiskProperties == nil {
```
rge-pure, line 374: According to the documentation of Azure -  If creationData.createOption is Empty, this field is mandatory and it indicates the size of the disk to create.

Also, this parameter caused panic before - https://portworx.atlassian.net/browse/PWX-29069

PR #141, opened 2023-03-28
PWX-30093: OCP bring up issue with vSphere

nrevanna:
vSphere client is being used in some paths which were being ignored
      thinking it's not needed. Reenabling them

Closes #

Testing:
Ran the fix on the cluster provided by Santhosh and px came up.

### DIFF
--- a/vsphere/vsphere.go
+++ b/vsphere/vsphere.go
@@ -2,7 +2,6 @@ package vsphere
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


vsphere/vsphere.go
```diff
@@ -767,10 +764,16 @@ func (ops *vsphereOps) getDatastoreToUseInStoragePod(
 	}
 
 	spec.DeviceChange = deviceChange
-	// It's best effort locking. If we didn't get a lock no harm done.
-	storeLock, lockErr := ops.dsLock.LockWithKey(ops.cfg.VMUUID, storagePod.Name())
+	var (
+		storeLock *store.Lock
+		lockErr error
+	)
+	if ops.dsLock != nil {
```
adityadani, line 771: Should we explicitly error out if dsLock is nil? 
I know that there is no code that invokes this function without initializing store. I am trying to future proof it to ensure `recommendDatastore` is always called with store initialized.

adityadani (approved):

PR #142, opened 2023-03-28
PWX-30091: Putting all new keys under the storePrefix

nrevanna:
Previously, all keys were put under the domain name
    e.g pwx/<cluster-id>, all keys would go under it.
    This could cause collisions since we allow creating new Stores
    Newly created stores will have all their keys under the domain name
    pwx/<cluster-id>/<store-name>/<keys>

Closes #


store/store_kv.go
```diff
@@ -84,20 +88,32 @@ func (kv *kvStore) Unlock(storeLock *Lock) error {
...
 		return nil, err
 	}
+
 	kvPair.lockedWithKey = true
 	return kvPair, err
 }
 
-func (kv *kvStore) lockWithKeyHelper(owner, key string) (*Lock, error) {
+func (kv *kvStore) lockWithKeyHelper(owner, key string, lockWithkey bool) (*Lock, error) {
+	if lockWithkey {
```
adityadani, line 114: nit: can we just move this logic of creating a path to individual callers, so that it is explicit what key they need to lock on?  Similar to how you have in Lock() API. 
This makes it easier to understand what key will eventually be locked

adityadani (approved):

store/store_kv.go
nrevanna: Addressed

PR #144, opened 2023-03-31
[PWX-30222] cloud provider unit testing example

alicelyu-px:
https://portworx.atlassian.net/browse/PWX-30222


pp511:
Can we run a sanity of eks with cloud drives to make sure ec2Wrapper client is working?

alicelyu-px:
> Can we run a sanity of eks with cloud drives to make sure ec2Wrapper client is working?

vendored in this change at https://github.com/portworx/porx/pull/11315
test automation passed at https://jenkins.pwx.dev.purestorage.com/job/Torpedo/job/eks-pool-expand-v1/61/

aws/aws_test.go
```diff
@@ -104,3 +105,44 @@ func TestAwsGetPrefixFromRootDeviceName(t *testing.T) {
...
 }
+
+type mockEC2Client struct {
+	ec2iface.EC2API
+	Vol *ec2.Volume
+}
+
+func (m mockEC2Client) CreateVolume(*ec2.CreateVolumeInput) (*ec2.Volume, error) {
+	return m.Vol, nil
+}
```
pp511, line 116: I am not sure what is the purpose of this function here. Ca you explain how is it being used in the test?

aws/aws_test.go
```diff
@@ -104,3 +105,44 @@ func TestAwsGetPrefixFromRootDeviceName(t *testing.T) {
...
+		expectedRes, _ := c.volResult.(*ec2.Volume)
+		s := &awsOps{
+			ec2: &ec2Wrapper{
+				Client: mockEC2Client{Vol: expectedRes},
+			},
+		}
+		res, err := s.Create(c.volumeTemplate, nil, nil)
+		if c.expectedErr.Error() != err.Error() {
+			t.Errorf("%s failed, expected error: \n%v\n but got: \n%v", c.name, c.expectedErr, err)
+		}
```
pp511, line 143: nit.. I think we should use assert.Equal
Something like assert.Equal(t, c.expectedErr, err, "<msg>")

aws/aws_test.go
alicelyu-px: This is to mock the return value of this interface function. 
type mockEC2Client struct {
	ec2iface.EC2API
	Vol *ec2.Volume
}
CreateVolume() will return what we defined as Vol instead of executing the function that involves cloud provider api calls.

pp511: Got it. My bad.

aws/aws_test.go
```diff
@@ -104,3 +105,44 @@ func TestAwsGetPrefixFromRootDeviceName(t *testing.T) {
...
+		if res != c.volResult {
+			t.Errorf("%s failed, expected result %s but got %s", c.name, c.volResult, res)
+		}
```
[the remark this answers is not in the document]
pp511: Instead of returning nil and comparing against it, should we return a result that simulates an actual response and compare against it.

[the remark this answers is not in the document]
alicelyu-px: I'm only providing the most simple test case in this PR for the purpose of demonstrating a way to mock cloud provider call. This is by no means a comprehensive unit test for the function to test. After this is merged, we can add more test cases and more tests in future.

pp511 (approved):
Looks good to me.
It'd be better to get a review from Naveen too before merge.

nrevanna (approved):

nrevanna (approved):

PR #145, opened 2023-04-04
PWX-23110: use go.etcd.io import instead of coreos/etcd

pureneelesh:
As part of move to etcd 3.4.24, we need to use go.etcd.io import everywhere.

Replaced github.com/coreos/etcd by go.etcd.io/etcd.

Also, added sched-ops to replace section to avoid error in go mod download.

PWX-23110


nrevanna (commented):
I don't see any updates to `vendor/github.com/portworx/kvdb`.
vendor/github.com/portworx/kvdb/Gopkg.lock references `github.com/coreos/etcd`. Will that be a problem?

nrevanna (approved):

pureneelesh:
> I don't see any updates to `vendor/github.com/portworx/kvdb`. vendor/github.com/portworx/kvdb/Gopkg.lock references `github.com/coreos/etcd`. Will that be a problem?

Vendored in the latest kvdb.

PR #146, opened 2023-04-07
Added pagination support to volume Enumerate() API of oracle cloud.

vinayakshnd:
Enumerate() API needs to list all the block volumes. For oracle, currently this was not happening as support for pagination was missing. When there are more than 200 volumes in users oracle account's compartment, pagination is required.
Closes #

### DIFF
--- a/oracle/oracle.go
+++ b/oracle/oracle.go
@@ -1000,34 +1000,50 @@ func (o *oracleOps) Enumerate(volumeIds []*string,
 	for _, volIds := range volumeIds {
 		volIDsMap[*volIds] = *volIds
 	}
-	for _, vol := range resp.Items {
-		_, ok := volIDsMap[*vol.Id]
-		if !ok {
-			continue
-		}
+	for {
+		for _, vol := range resp.Items {
+			_, ok := volIDsMap[*vol.Id]
+			if !ok {
+				continue
+			}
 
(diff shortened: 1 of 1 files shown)


oracle/oracle.go
```diff
@@ -1000,35 +1000,51 @@ func (o *oracleOps) Enumerate(volumeIds []*string,
 	for _, volIds := range volumeIds {
 		volIDsMap[*volIds] = *volIds
 	}
-	for _, vol := range resp.Items {
-		_, ok := volIDsMap[*vol.Id]
-		if !ok {
-			continue
-		}
+	for {
```
piyush-nimbalkar, line 1003: nit: You can make the first ListVolumes call also inside the for loop. A slightly optimized way to write this could be - 
```
var nextPage *string
for {
        req := core.ListVolumesRequest{
		CompartmentId: common.String(o.compartmentID),
	}
        if nextPage != nil {
                 req.Page = nextPage
        }
        resp, err := o.storage.ListVolumes(context.Background(), req)
	if err != nil {
		return nil, err
	}
        
        // process

        if resp.OpcNextPage == nil {
                  break
        }
        nextPage = resp.OpcNextPage
}

```

piyush-nimbalkar (approved):

PR #147, opened 2023-06-02
[PWX-31572] make FreeDevice() interface call available for px

alicelyu-px:
Needed by porx PR: https://github.com/portworx/porx/pull/11493


nrevanna (approved):
LGTM

PR #148, opened 2023-06-28
PWX-32045: Pass IOPS during pool expand for gp3 drives

nrevanna:

Closes #

Testing details here: https://github.com/portworx/porx/pull/11613

### DIFF
--- a/aws/storagemanager/aws.go
+++ b/aws/storagemanager/aws.go
@@ -83,7 +83,7 @@ func determineIOPSForPool(instStorage *cloudops.StoragePoolSpec, row *cloudops.S
 	if instStorage.DriveType == DriveTypeGp2 {
 		return instStorage.DriveCapacityGiB * Gp2IopsMultiplier
 	} else if instStorage.DriveType == DriveTypeIo1 || instStorage.DriveType == DriveTypeGp3 {
-		// For io1 volumes we need to specify the requested iops as the provisioned iops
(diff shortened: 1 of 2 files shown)


aws/storagemanager/aws.go
```diff
@@ -82,7 +82,7 @@ func (a *awsStorageManager) RecommendStoragePoolUpdate(
 func determineIOPSForPool(instStorage *cloudops.StoragePoolSpec, row *cloudops.StorageDecisionMatrixRow, currentIOPS uint64) uint64 {
 	if instStorage.DriveType == DriveTypeGp2 {
 		return instStorage.DriveCapacityGiB * Gp2IopsMultiplier
-	} else if instStorage.DriveType == DriveTypeIo1 {
+	} else if instStorage.DriveType == DriveTypeIo1 || instStorage.DriveType == DriveTypeGp3 {
```
harsh-px, line 85: how about io2 types?

If it's low hanging, let's add a UT. Okay to skip here as this is a trivial function.

nrevanna: We don't support io2 today.
```
  drive_type: gp2
  drive_type: gp3
  drive_type: io1

```
We have unit tests in aws_test.go. It seem to have caught one scenario. I am testing that on an actual cluster.

nrevanna: Added a unit test. The failure was a test parameter failure not a product issue.

harsh-px (approved):

PR #163, opened 2024-04-18
PWX-36933 - update decision matrix for new disk types

dtalreja-px:
Update decision matrix for azure disks - premiumv2 and ultradisk

PWX-36933

Teted -
PASS
coverage: 90.9% of statements
ok  	github.com/libopenstorage/cloudops/vsphere/storagemanager	0.010s	coverage: 90.9% of statements

### DIFF
--- a/specs/decisionmatrix/azure.yaml
+++ b/specs/decisionmatrix/azure.yaml
@@ -219,222 +219,24 @@ rows:
           priority: 3
           thin_provisioning: false
           drive_type: "Standard_LRS"
-        - min_iops: 120
-          max_iops: 120
+        - min_iops: 19000
+          max_iops: 80000
           instance_type: "*"
           instance_max_drives: 8
           instance_min_drives: 1
           region: "*"
           min_size: 32
-          max_size: 64
+          max_size: 65536
           priority: 2
           thin_provisioning: false
           drive_type: "PremiumV2_LRS"
-        - min_iops: 240
-          max_iops: 240
-          instance_type: "*"
-          instance_max_drives: 8
-          instance_min_drives: 1
-          region: "*"
-          min_size: 64
-          max_size: 128
-          priority: 2
-          thin_provisioning: false
-          drive_type: "PremiumV2_LRS"
-        - min_iops: 500
-          max_iops: 500
-          instance_type: "*"
-          instance_max_drives: 8
-          instance_min_drives: 1
-          region: "*"
-          min_size: 128
-          max_size: 256
-          priority: 2
-          thin_provisioning: false
-          drive_type: "PremiumV2_LRS"
-        - min_iops: 1100
-          max_iops: 1100
-          instance_type: "*"
-          instance_max_drives: 8
-          instance_min_drives: 1
-          region: "*"
-          min_size: 256
-          max_size: 8192
-          priority: 2
-          thin_provisioning: false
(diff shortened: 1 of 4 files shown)


pnookala-px (approved):

specs/decisionmatrix/azure.yaml
```diff
@@ -219,222 +219,24 @@ rows:
...
-          thin_provisioning: false
-          drive_type: "UltraSSD_LRS"
-        - min_iops: 20000
-          max_iops: 20000
-          instance_type: "*"
-          instance_max_drives: 8
-          instance_min_drives: 1
-          region: "*"
-          min_size: 32768
           max_size: 32768
```
nrevanna, line 240: Ultra disk's capacity can technically go up to 64T. It might need additional approval from Azure support, but it can. We should keep out size at 64T as well so we don't block a customer when they have the necessary permission. Azure API will anyway return an error if they don't have that permission.
https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disk-size

specs/decisionmatrix/azure.yaml
```diff
@@ -219,222 +219,24 @@ rows:
           priority: 3
           thin_provisioning: false
           drive_type: "Standard_LRS"
-        - min_iops: 120
-          max_iops: 120
+        - min_iops: 19000
```
nrevanna, line 222: Why is this 19000?
Per math here
https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#premium-ssd-v2-iops
formula will be
((size - 6) * 500) + 3000
((32-6) * 500) + 3000
= 16000.  
What am I missing?

nrevanna:
https://github.com/libopenstorage/cloudops/blob/PWX-36933/azure/storagemanager/testspecs/azure-storage-decision-matrix.yaml
Needs an update too?

specs/decisionmatrix/azure.yaml
dtalreja-px: **Now that I read it again, I feel minimum should be 3000 iops** . (All Premium SSD v2 disks have a baseline IOPS of 3000 that is free of charge. After 6 GiB, the maximum IOPS a disk can have increases at a rate of 500 per GiB, up to 80,000 IOPS. So an 8 GiB disk can have up to 4,000 IOPS, and a 10 GiB can have up to 5,000 IOPS. To be able to set 80,000 IOPS on a disk, that disk must have at least 160 GiBs. Increasing your IOPS beyond 3000 increases the price of your disk)
@nrevanna for a 32 gb disk - minimum can still be 3000 and maximum can be 16000.
should I go with 3000 ??

dtalreja-px: @nrevanna can you also help me with the priority here ?? If you know on what basis this is set ?

nrevanna: Priority is not used today. You can keep the same value as other disk type entries
3000 makes sense. Let's keep that for min iops.

specs/decisionmatrix/azure.yaml
dtalreja-px: updated

specs/decisionmatrix/azure.yaml
dtalreja-px: fixed.

PR #164, opened 2024-05-03
PWX-37022 Support resize-drive for ultradisk

dtalreja-px:
We were unable to resize UltraDisk because for ultra disk the minimum IOPS change with every GiB.
```The minimum guaranteed IOPS per disk are 1 IOPS/GiB, with an overall baseline minimum of 100 IOPS. For example, if you provisioned a 4-GiB ultra disk, the minimum IOPS for that disk is 100, instead of four.```
https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disk-iops

Closes https://purestorage.atlassian.net/browse/PWX-37022

In the current implementation, we did not update the minimum readonly IOPS andminimum ReadwriteIOPS. For ultra as the minimum changes with every GB - the resize was failing.

Automation - https://jenkins.pwx.dev.purestorage.com/job/Users/job/Dolly/job/dtalreja-Azure-ultraDisk/12/

### DIFF
--- a/azure/azure.go
+++ b/azure/azure.go
@@ -613,21 +613,21 @@ func (a *azureOps) Expand(
 	// Azure resizes in chunks of GiB even if the disk properties variable is DiskSizeGB
 	newSizeInGiBInt32 := int32(newSizeInGiB)
 	disk.DiskProperties.DiskSizeGB = &newSizeInGiBInt32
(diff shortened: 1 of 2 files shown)


vinayakshnd (approved):
LGTM

